PR #10, opened 2015-12-08
Add an autotools system to enable easier packaging

ikeydoherty:
This makes it far easier for us to manage and distribute the icon theme, as it's appearing in quite a few repos :)

If/when this is merged, remember to edit the version number in `configure.ac` to reflect the current version of the package, then git add/git commit that change, and pop a git tag in for this version.

With that done, you can actually generate distributable source tarballs for elementary-icon-theme as follows:

```
./autogen.sh
make distcheck
```

GitHub allows you to edit the tags after pushing them, so you can actually upload this `elementary-icon-theme-$VERSION.tar.xz` for people to use as the formally distributed tarball. This ensures its complete, and usable in building packages.

Cheers :)

N.B I decided to recreate the `@2x` symlinks during the autotools install, much cleaner this way, and doesn't bork tarballs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+Makefile.in
+aclocal.m4
+autom4te.cache/
+configure
+install-sh
+missing
+Makefile
+config.log
+config.status
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,0 +1,33 @@
+ICON_DIRS = \
+	actions \
+	mimes \
+	categories \
+	devices \
+	emotes \
+	cursors \
+	places \
+	status \
+	emblems \
+	apps
+
+EXTRA_DIST = \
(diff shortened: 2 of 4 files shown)


danrabbit:
My gut reaction is that if we were to add a build system it should probably be CMake. We don't really use autotools. I'll let Cody make the call though about what he wants to do

danrabbit:
Not sure if Cody contacted you in Slack, but it sounds like he wants to write a CMake system

ikeydoherty:
He didn't and you know that, I only saw the message context on slack by searching back.
Please let people know when you have a suitable build system in place, I won't put anything into our repos that has a subpar/absent build system, so would like to know when one arrives. Thanks.

PR #15, opened 2015-12-09
drop dialog-close/cancel/no

danrabbit:
As far as I know, these are basically vestiges from the days when dialog buttons had icons. Pretty sure it's safe to drop them


snwh:
I concur

PR #47, opened 2015-12-19
moved icons; made symlinks

snwh:
A little bit of reorganization.


danrabbit:
Please use git mv for the camera icons so it records as a move and not a new file

snwh:
As suspected, it makes no difference to use ```git mv``` see other pull request #48

PR #52, opened 2015-12-20
change purple

danrabbit:


snwh:
- [ ] stroke on the 128 mail-reply-all is a bit heavy
- [ ] missed new purple for 48 object-rotate-* icons

PR #57, opened 2016-01-01
Optimized (~68% size) icons

lucasmciruzzi:
Optimized icons (using svgo).

The total weight of the icons is about 18MB, and with this optimization is a little below 6MB.


danrabbit:
Hey thanks for your branch. This is going to take a while to review haha

embik:
This branch is obviously outdated by now, but what's about the optimizations applied here?

lucasmciruzzi:
@embik I could code a script to apply the svgo optimizations in all the files, but that would add a node.js dependency that I think you don't want.

embik:
Maybe we can run svgo during `pre-commit` when it's installed and add a section about it in the readme. I wonder how @danrabbit feels about adding this optimization.

danrabbit:
At the moment, we do have a pre-commit to create plain SVGs and vacuum defs which has resulted in smaller file size over time. I am concerned a bit about using optimized SVG as both @snwh and I have had issues with thumbnailing before, which is pretty inconvenient. If there are additional size optimizations that we can add to the pre-commit and verify aren't destructive, I'm open to that :)

PR #80, opened 2016-03-19
added format icons for i18n

snwh:


danrabbit:
And you get i18n! And you get i18n! And you get i18n!

PR #82, opened 2016-04-10
New "less generic" software center icon

snwh:
Solves #1


snwh:
Hell, ignore this my script borked these icons.

PR #100, opened 2016-08-09
Added 48px microphone symbolics

snwh:


danrabbit:
- [ ] change fill to #bebebe to match other icons
- [ ] The x in muted needs `error` class in to show as red. Not sure this should be red though. The x in volume mute is not.

snwh:
@danrabbit 6px looked small, went with 7px.

danrabbit:
- [ ] combine paths to save space with definitions ;)

PR #108, opened 2016-08-23
Improved exclamation point motif

snwh:
Updates dialog-error and dialog-warning with an improved !


danrabbit:
can you revisit the 16px and especially 24px icons? On HiDPI the 24px ones are noticeably very square

danrabbit:
fixes #109

PR #111, opened 2016-08-23
fix porportional issues with calculator icons

### LINKED ISSUE #110: Calculator icons have odd padding
The calculator icons have uneven padding around some elements.

snwh:
Adjusts calculator icons by a pixel here and a pixel there in height/width. Fixes #110


danrabbit:
I'm not a fan of making them not-centered. It especially looks weird when it's in a list with other icons like in the applications menu search

PR #121, opened 2016-08-24
improved process-completed-symbolic

snwh:
I thought that it could use a nicer checkmark. Top: old, bottom: new.

screenshot from 2016-08-24 11-28-48


danrabbit:
I'm not sure I like this check more. It seems a little horizontal

PR #129, opened 2016-08-29
improved edit-delete icons

snwh:


danrabbit:
Hm, I think maybe what we actually wanted to do here was to symlink the rest of these to `user-trash`

snwh:
Well, I just wasted 10 hours doing all this 😭

PR #142, opened 2016-10-25
make the arrow less bulky and gross

danrabbit:


snwh:
real arrows have curves

PR #143, opened 2016-10-26
update app default icon to use gear shape from mimes

danrabbit:


danrabbit:
Sam wants the gear to be as big as the old one on 64px

PR #154, opened 2016-12-15
added rtl mimetype icons

snwh:


danrabbit:
Right on :) Couple things:

- [ ] I wonder if code mimes like scripts and make files should not have rtl variants. Afaict, putting Scratch into RTL (for example) doesn't change this. I think code is effectively written in English

- [ ]  We're missing `office-document-rtl`
- [ ]  Might want to double check symlinks. I noticed at least one missing one in 128/wordprocessing.svg

snwh:
@danrabbit 128/wordprocessing is still there.

PR #201, opened 2017-04-19
Make the bubble on notification-new-symbolic larger

danrabbit:


snwh:
Oh right, I was gonna do this. :joy:

PR #204, opened 2017-05-01
redesign 24px symbolic battery icons

danrabbit:
Change battery shape to match color icons and make use of that 24px canvas by widening a little

battery


snwh:
In the case of these symbolic icons, the addition of curving only achieves a reduction in clarity especially at 24px. I'd approve a wider icon though. :wink:

PR #209, opened 2017-05-03
remove lightning from fully charged state

danrabbit:
Getting some feedback that this icon is now too "loud". Previously we used the lightning with inverted colors, but I feel like this doesn't work as well now that the lightning is vertical and not fully contained in the level bar. I want to try removing the lightning image here and sticking with the green color for a couple reasons:

* The lightning indicates that charging is occurring. In this state, charging is no longer occurring we're just not discharging. So I think removing the lightning makes sense here.

* Green is our "success" state color and the real goal of this icon is to indicate that the charging task has been completed. Otherwise if we don't care about notifying of this completion, we should probably just symlink to the charging-full state


kurisubrooks:
Your work on these new symbolic icons are truly stunning. Can't wait to see them incorporated into eOS! 🙌

PR #210, opened 2017-05-03
Add System Monitor icon

### LINKED ISSUE #169: Missing utilities-system-monitor icon [$5]
It was removed?

danrabbit:
Fixes #169 
- [ ] 16
- [ ] 24
- [ ] 32
- [ ] 48
- [x] 64
- [x] 128


cassidyjames:
I thought we consciously removed this because we don't ship a system monitor, and so all our icon was doing was rebranding the GNOME app?

danrabbit:
Yeah but it's part of the fd.o icon spec

cassidyjames:
Do we want/need more sizes before we merge?

danrabbit:
Yeah, adding a checklist to description

waldyrious:
> the icon in its github repo is different and nowhere near as cool.

FWIW I actually find that one more pleasing, and also suspect that the increased contrast may make it work better at smaller sizes.

For convenience, here's the image from the repo:

screenshot from 2018-02-25 17-17-50

danrabbit:
We definitely can't steal an icon from a third party app. That would be basically robbing them of their branding.

PR #216, opened 2017-05-08
Symlink meson mime to make file mime

danrabbit:
This makes meson files look like build files instead of just text files


cassidyjames:
It looks like there were also some non-visible changes to the makefile mime SVGs; was that intentional?

danrabbit:
Fixed :)

PR #230, opened 2017-06-12
make changelog mime flat, symlink patch

danrabbit:
Featuring colors from our new neato palette shades thingy


snwh:
I'm not sure he was completely done, he only changed 4 icons :wink:

danrabbit:
Those are the only sizes that we carry right now :p I'll do another PR at some point with more sizes

PR #233, opened 2017-06-14
new speaker icons

### LINKED ISSUE #90: [$100] update audio-speakers icon
This is, at this point, one of the oldest icons in the set. It probably needs to be updated.

One of the big questions here is that the perspective is pretty different from other icons. Maybe we should consider a more straight-on design.

Another issue is if the speaker should remain "disembodied". Maybe it should have some kind of case? We should probably consider what kind of speakers people are connecting to their computers. Should it look like one of those hand-held bluetooth speakers or like something from a home audio setup? Should we stick to something that is more iconic even if it's less "realistic"?

snwh:
Updated speaker icons, closes #90


danrabbit:
screenshot from 2017-06-14 11 20 36

The 32px one is a little large on the canvas. If you compare it to `media-optical` for example it's a few pixels wider

PR #241, opened 2017-06-14
match symbol from color markup mime and add symlinks

danrabbit:
It doesn't look like text-x-html is actually a real thing...


mimes/symbolic/text-htmlh-symbolic.svg
```diff
@@ -0,0 +1 @@
+text-html-symbolic.svg
```
cassidyjames: Is this symlink name typo'd?

cassidyjames: Never mind, TIL HTMLH is a thing I guess?

danrabbit: Yeah I'm not 100% where these mime names are all from. It might be bullshit. Just trying to match the color names

PR #242, opened 2017-06-14
make audio-x-generic more generic

danrabbit:
Proposing for feedback

sound1

My thinking is that the current audio mime leans heavily towards music, but we can't know the contents of the audio file. So it might be better to use a mime that more conveys raw audio than music specifically

### DIFF
--- a/mimes/128/audio-x-generic.svg
+++ b/mimes/128/audio-x-generic.svg
@@ -9,246 +9,192 @@
    version="1.1"
    width="128"
    height="128"
-   id="svg3971">
+   id="svg3172">
   <defs
-     id="defs3973">
+     id="defs3174">
     <linearGradient
-       x1="16.626165"
-       y1="15.298182"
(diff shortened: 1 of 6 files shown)


snwh:
I'm not going to disagree completely. But "leans heavily" is a bit much, I think the music note for audio files is so widely used that changing it seems a bit unnecessary and more often than not for most people audio files are music.

Also I think that wave design is too realistic. Something simpler and symbolic may work better.

cassidyjames:
I'm a fan of the more generic design; it's always a bit weird to have sound recordings or audiobooks look like music, and music looking like sound waves is still appropriate.

danrabbit:
Yeah on second thought, the wave form could be difficult to do for the symbolic variant as well. Looking at the symbolic icon, we're currently symlinking the eighth notes icon. So we could at least get higher contrast and ditch the album metaphor

sound

PR #243, opened 2017-06-15
new microphone icons

### LINKED ISSUE #238: Improve microphone icons
These are some of the oldest in the set, could use a new design.

snwh:
Closes #238


danrabbit:
Copying from Slack:

> do you think we can make the mic a little less stocky? Comparing with the symbolic version, especially at the smaller sizes, the stand seems to take up more of the canvas than the mic itself

screenshot from 2017-06-15 09 57 59

Left: original, center: symbolic, right: taller mic, smaller stand

PR #247, opened 2017-06-16
drop aptdaemon icons

danrabbit:
We don't use aptdaemon anymore because we don't use Ubuntu's updater anymore


snwh:
long overdue

PR #253, opened 2017-06-18
text-css for all sizes, colors to new palette

danrabbit:
Stole the symbol from the Code icon


snwh:
seems a bit small on the 128

danrabbit:
It's the same size as the 64px one relative to the background :man_shrugging:

snwh:
Mathematically yes, but they're not as tall on the bg as some of the other symbols so we have to make up for that optical illusion, I feel, by making it slightly larger.

PR #262, opened 2017-07-18
refactor weather symbolic icons

snwh:


snwh:
screenshot from 2017-07-18 16 13 09
screenshot from 2017-07-18 16 13 22

PR #270, opened 2017-07-19
overhauled device icons

snwh:
- removed non-standard icons
 - redesigned legacy gnome icons
 - redesigned our icons to be more consistent
 - minified


snwh:
will do

PR #271, opened 2017-07-19
update drive icons

snwh:
I axed the perspective because they are symbolic icons and should be flat in my view –the drive icons are the only in the set with a non-flat perspective


danrabbit:
One thing I really don't like about flattening is that the symbolic icon now looks very different from its color counterpart. This is especially clear in context:

screenshot from 2017-07-18 18 12 00

I'd prefer these icons to be much closer

snwh:
Fair point, I could make it look more like the coloured version.

PR #291, opened 2017-07-21
added symbolic emotes set

snwh:


danrabbit:
What do you think about inverting the color on these symbolic emotes to more closely match the values in the color set. Something in the vein of https://dribbble.com/shots/1872329-Sentoj

I think that would keep more closely with the changes we recently made to symbolic devices and especially for emotes like `face-cool` I think it would look more natural

PR #303, opened 2017-07-24
update remaining action icons

snwh:
- update insert-text, font-select and spell-check
 - de-gnome call icons
 - make process-stop not look like window-close
 - use symlinks
 - remove old panel icon system-devices
 - minification


danrabbit:
I think we should probably update the direction of the slashie on the ban symbol while we're here no?

snwh:
@danrabbit done

PR #306, opened 2017-07-24
updated few categories 48 symbolic icon

snwh:


snwh:
@danrabbit but puts it in proportion with the rest of the 48px symbolics. The other symbolic icons at this size aren't exactly the size of their coloured counterparts, they're closer to the 32 version on a 48 canvas

danrabbit:
It looks like the music icon moved up a few pixels. Is that intentional?

snwh:
@danrabbit yeah, it looked really low to me

PR #312, opened 2017-07-24
updated import export action icons

snwh:


danrabbit:
Looks like your last commit just reverts your direction fixes

danrabbit:
Looks like these are not the same size/shape as their 24px counterparts

screenshot from 2017-07-25 10 01 57

PR #317, opened 2017-07-24
updated wired icons

snwh:


snwh:
@danrabbit I put it in the same position as it is on network-wired-secure

PR #324, opened 2017-07-26
Add mail-important-symbolic status icon

danrabbit:


snwh:
already added as `mail-mark-important` 😉

danrabbit:
I wanted to add a separate name because one is an action and another is a status. So a developer might want to use mail-mark-important in a toolbar or button and I don’t think we’d want it to be colored in that case. But we would want it to be colored in a list view for example

snwh:
Fair enough

PR #333, opened 2017-08-02
adjusted 16px mail place icons

snwh:
- a little less chunky of an icon
- make them like the symbolic variant


danrabbit:
I think I'm -1 on getting rid of the perspective here

snwh:
Mail's icon doesn't have perspective

PR #342, opened 2017-09-18
Add secure cellular icons

danrabbit:
for VPN


davidmhewitt:
Can confirm that these look good in the indicator :+1:

PR #343, opened 2017-09-18
24 process stop symbolic

### LINKED ISSUE #334: Missing 24px process-stop
You can see it in the headerbar for Epiphany. It looks bad :P

danrabbit:
Fixes #334 
Restores the old 24px process-stop-symbolic icon just to avoid breaking epiphany


codygarver:
I think this icon is also used in Files progress bar, I noticed it was broken lately.

PR #376, opened 2017-12-11
Night light icons

danrabbit:
Proposed icons for Night Light.


Philip-Scott:
On the latest symbolic icons, one of them is a bit off-centered from the others. You can tell if you toggle the indicator while opened, the "rays" are not aligned :)

snwh:
Any reason why central the circle for the 24px version is off the pixel grid?

danrabbit:
@snwh because the little sun rays are only 1 x 2px so centering would put them off the grid

snwh:
@danrabbit yes, but the inner part of the circle is off the grid as a result

PR #393, opened 2018-01-17
Redo edit flag symbolic

danrabbit:


cassidyjames:
I like the style, but I worry that the thinness might be too lightweight in use.

cassidyjames:
But I see that it's for consistency with the emoji… :man_shrugging:

PR #400, opened 2018-01-18
Make symbolic time icons consistent

### LINKED ISSUE #787: Clock and timer icons different weights
I was prototyping something in Parental Controls, and it's strange that these icons have different weights at the same size.

image

danrabbit:
Fixes #787 

This rebases all of the time-related symbolic icons on a 14px circle with a 1px stroke width

The arrow in document-open-recent is now consistent with the rotation lock icons

The hands are now all pointing to the same time


danrabbit:
The reason I went with this arm position is to work around a couple problems:

One big one is that since we use a canvas with an even number of pixels, centering lines would put them on half pixels. The old alarm icon works around this by shifting the whole icon on the canvas down and left by a pixel. So whatever arm position we choose shouldn't be at any of the 15 minute positions

Another problem is that we have badges in the bottom right corner of some icons and I'm trying to avoid overlapping those (as the position in the old document-open-recent does).

jeremypw:
@danrabbit: Has conflicts, is blocked and has no recent activity.  Does this need closing or converting to draft?

danrabbit:
Yeah I'm just gonna close it

PR #409, opened 2018-01-26
Add symlinks for security-foo icons

danrabbit:
symlink all the things


TraumaD:
I believe this is no longer needed if #419 is accepted.

PR #443, opened 2018-03-02
Improve the symbolic open-menu gear

danrabbit:
## Before

screenshot from 2018-03-02 13 51 02

## After

screenshot from 2018-03-02 14 07 03


Shnatsel:
Oh yeah that's much hotter. Maybe still a little too heavy in this context though.

PR #450, opened 2018-03-13
Bring some blue and green icons to the palette, also 128px chat icon

micahilbery:


micahilbery:
I'm not 100% sold on the green gradient in the browser icon. It just felt kinda dark and I tried to lighten it up. Thoughts?

waldyrious:
I was under the impression that the design language for the icons was being consolidated towards flatter color areas and subtler gradients, yet this change (aside from the color normalization) seems to go in the opposite direction. Am I missing something?

PR #452, opened 2018-03-13
Color folders

danrabbit:
Opening a PR for comments ;)


danrabbit:
I think that default folder color might be too bright

screenshot from 2018-03-13 10 40 35 2x

micahilbery:
I agree. It gets totally blown out when on hover.

micahilbery:
I changed the defaults to blue for better contrast against sidebars and changed photos folder silver like camera lens.

danrabbit:
The photo glyph is pretty off-center now ;)

waldyrious:
Re: 7ad7e5d (and speaking as an end user) I have no issue with reverting the default folder icon to blue, but IMHO it looks weird that all predefined folders are colored, and precisely the pictures folder is kind of "greyed out". How about using a light pink for it?

fabiozaramella:
I **love** this set of colors, just a note: the yellow seems a bit off, looks darker when compared to other colors. 

schermata del 2018-03-13 23 54 31

aljelly:
For the new colors on the special folder icons, do you think they could be potentially confused with color tags on folders? Considering that the colors for color tags are now consistent with the palette.

My opinion: I like that the default folder icon has been updated (in regards to changing it to an updated blue). I'm not so sure about the colors on the special folder icons. However I think I would need to try the branch with the new icons first in order to better assess.

micahilbery:
@aljelly I don't think it will be confusing with the tags because usually you use those on subfolders for sorting not in the home folder which is already sorted with custom icons. And as far as the smaller sizes go, yeah we will update those if we decide to push further with the concept.

cassidyjames:
@micahilbery could you update this PR since we have the manilla folders now? I still think there's value in the colored folders for special folders, especially with their propagation to the media apps and whatnot.

micahilbery:
@cassidyjames Yeah, I'll try and start back up on it tomorrow if I get a chance.

micahilbery:
This time I only focused on the folders that appear in the home folder and in applications with an associated file type.

Issue #457, opened 2018-03-14
preferences-desktop-keyboard 64px is missing


fabiozaramella:
Now that shortcut-overlay uses this icon we should ship it.

This is the appearance when you set plank icons to maximum size:
![schermata da 2018-03-14 21-05-45](https://user-images.githubusercontent.com/5459945/37428270-fcbcb40a-27cb-11e8-8c85-2a79a99f4172.png)

PR #460, opened 2018-03-15
add 64px keyboard prefs

danrabbit:
Fixes #460 

While we're here, let's use palette colors


waldyrious:
> Fixes #460

You probably meant #457 :)

PR #462, opened 2018-03-19
Palette blue folders

micahilbery:
If we decide against #452 this is another option to bring the current folders up to the new palette.


micahilbery:
Context:
files

waldyrious:
For reference, this was superseded by #634.

PR #474, opened 2018-04-13
Convert go-* action icons to the new colorscheme

ochosi:
There are still some arrows using the "old greens". This patch converts them.

### DIFF
--- a/actions/48/go-bottom.svg
+++ b/actions/48/go-bottom.svg
@@ -6,100 +6,110 @@
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:xlink="http://www.w3.org/1999/xlink"
-   version="1.1"
-   width="48"
-   height="48"
-   id="svg2510">
+   width="48px"
+   height="48px"
+   id="svg2510"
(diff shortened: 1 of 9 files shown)


waldyrious:
@ochosi the `actions/64/go-down.svg` file has a different gradient color than the other arrows. I suppose that was an overlook?

waldyrious:
Ah, makes sense then. I actually only compared the visual diffs, and the difference in the color difference seemed quite a bit more marked in the new gradients than in the old ones, which led me to assume the latter were consistent. Here's a screenshot that hopefully illustrates better what I mean:

screenshot from 2018-04-14 13-15-14

That said, it looks like the go-* arrows only exist in colored variants in the 48px and 64px sizes, and in particular the 64px size only has the go-down arrow. Since the colors are already being changed here, and there aren't really two sets of gradient definitions, but a single outlier, I would suggest directly addressing the inconsistency in this PR. Of course, that's just a suggestion to solve two issues at once -- I'll defer to the active contributors' judgment on this :)

ochosi:
Yes, and I agree. Also wanted to get feedback (and was also wondering what the 64px icon was actually used for)

PR #516, opened 2018-05-10
Make grape palette more like icons

micahilbery:
grape


danrabbit:
Tweaked these just a little bit, especially in the brighter shades

screenshot from 2018-05-10 09 42 49 2x

PR #536, opened 2018-05-13
Add 24 and 48 document-send

danrabbit:
Here's my proposal for `document-send`.

Cloud because this is different from `document-export`. This is a "put my file through the internet tubes" action.

Purple to match `document-import` because the file is going into the cloud


micahilbery:
:+1: Looks good to me! Assuming the use case is something similar to an upload button I definitely think this makes sense.

PR #559, opened 2018-05-15
mimes: text-x-ssa missing sizes

NikoKrause:
screenshot-area-2018-05-15-151514


NikoKrause:
I think the size for 64 is a little bit too small, but I compared to video-x-generic icons:
screenshot-area-2018-05-15-151730

PR #562, opened 2018-05-16
Add sizes for event-new

danrabbit:
Adds missing sizes for the action event-new


waldyrious:
Is it intentional that the second-to-last horizontal line is thicker than the other ones? It looks odd IMO, and the distinction is too subtle to appear unambiguously deliberate.

waldyrious:
Thanks for the explanation. Hopefully it isn't too conspicuous.

PR #570, opened 2018-05-16
add ruby mime

NikoKrause:


waldyrious:
Would a different style fit better with the other icons' aesthetic? For example, something similar to these:

blue_diamond_symbol svg
200px-noto_emoji_oreo_1f48e svg
twemoji_1f48e svg
sketch_logo_frame svg

The one in this PR feels excessively flat IMHO:

screenshot from 2018-05-17 11-47-44

By the way, all the images above are freely licensed on [Wikimedia Commons](https://commons.wikimedia.org/wiki/Category:Gemstone_icons).

NikoKrause:
Rubys official logo looks like this:
Ruby

waldyrious:
Yes, that could be used too, but I fear it might be too shiny to fit in with the rest of the icons.

waldyrious:
@danrabbit of the examples above, does any of them look like something that could be adapted to an improvement on the current design?

danrabbit:
I'd like to stick as close as possible to the official logo as possible. It looks like they have a symbolic version of the logo on the download button on their website. We could try that and see how it looks? https://www.ruby-lang.org/en/

Issue #582, opened 2018-05-17
mimes: inconsistent border colors


NikoKrause:
I noticed, that some mimetypes have different border colors compared to each other. This is more visible with dark themes.

Example with 48px icons:
![z-mime-48](https://user-images.githubusercontent.com/8415124/40193686-52b66720-5a08-11e8-92b4-68ba4df8138e.png)

### 16px
- [x] application-x-cd-image
- [x] application-octet-stream
- [ ] payment-cards* (?)
- [x] office-document
- [x] office-document-rtl
- [x] unknown
- [x] text-x-readme
- [x] text-x-authors
- [x] x-office-document-template
- [x] x-office-document-template-rtl
- [x] x-office-drawing-template

### 24px
- [x] image-x-generic (?)
- [x] application-vnd.debian.binary-package
- [x] office-document
- [x] office-document-rtl
- [x] x-office-document-template
- [x] x-office-document-template-rtl
- [x] text-x-authors
- [x] text-x-generic-template
- [x] text-x-generic-template-rtl

### 32px
- [x] application-x-cd-image
- [x] application-x-flash-video
- [x] model
- [x] office-document
- [x] office-document-rtl
- [x] x-office-document-template
- [x] x-office-document-template-rtl
- [x] text-x-install
- [x] text-x-generic-template
- [x] text-x-generic-template-rtl

### 48px
- [x] application-x-flash-video
- [x] office-document
- [x] office-document-rtl
- [x] x-office-document-template
- [x] x-office-document-template-rtl
- [x] x-office-drawing-template

### 64px
- [x] text-x-changelog
- [x] x-office-document-template
- [x] x-office-document-template-rtl

### 128px
- [x] application-vnd.debian.binary-package
- [x] application-x-cd-image
- [x] text-x-readme
- [x] text-x-script
- [x] unknown

NikoKrause:
Actually if you look with a dark theme at the 128px icons all of those have a lighter border than their smaller counterparts, except `application-vnd.debian.binary-package`

This one is consistent in 128:
![screenshot-area-2018-05-17-210350](https://user-images.githubusercontent.com/8415124/40198165-e3c084b4-5a15-11e8-977b-1982107e6a6c.png)

This one is inconsistent in 128 (and almost all 128px icons look like this one below):
![screenshot-area-2018-05-17-210411](https://user-images.githubusercontent.com/8415124/40198166-e4d660ee-5a15-11e8-84da-d6330b7991f0.png)

PR #592, opened 2018-05-19
Fix borders for 128px mimes

danrabbit:


waldyrious:
Does this address #582?

PR #600, opened 2018-05-29
Fix black square for harddisk symbolic

### LINKED ISSUE #599: Black square for 48px drive-harddisk
screenshot from 2018-05-28 20 48 00 2x

danirabbit:
Fixes #599


waldyrious:
For future reference, this PR appears to solve a similar problem as #480.

PR #634, opened 2018-08-12
Manila folders

danrabbit:
screenshot from 2018-08-12 14 21 55 2x


cassidyjames:
I like this, as it gives us a good base to build off of for other UI color schemes  (https://github.com/elementary/stylesheet/issues/375) and/or pulling in semantic colors for folders (https://github.com/elementary/icons/pull/452).

PR #635, opened 2018-08-13
Better 32px folder glyphs

danrabbit:
Makes the glyphs on the 32px folder icons look a bit nicer (and more proportionate to their 64px counterparts)


snwh:
@danrabbit did  you miss `folder-publicshare-open`?

danrabbit:
@snwh I was going to ask you, but it looks like it's the same size glyph already

Issue #667, opened 2018-09-24
Wired internet icons don't match between colored and symbolic


danrabbit:
For the colored icon we use what looks like an RJ45 port but for the symbolic icon we use three connected boxes/nodes/computers.

Something to consider is that it looks like tethered USB connections also use this namespace, so it's probably not accurate to use RJ45 to represent a wired connection

This covers:

Status/symbolic & 24
- [ ] network-error-symbolic
- [ ] network-idle-symbolic
- [ ] network-offline-symbolic
- [ ] network-receive-symbolic
- [ ] network-transmit-symbolic
- [ ] network-transmit-receive-symbolic
- [ ] network-vpn-acquiring-symbolic
- [ ] network-vpn-symbolic
- [ ] network-wired-acquiring-symbolic
- [ ] network-wired-disconnected-symbolic
- [ ] network-wired-no-route-symbolic
- [ ] network-wired-offline-symbolic
- [ ] network-wired-secure-symbolic
- [ ] network-wired-symbolic

status/48
- [ ] network-wired-disconnected
- [ ] notification-network-ethernet-disconnected
- [ ] notification-network-wired

PR #668, opened 2018-09-24
Made Wired network icon look like a RJ45 port to match color versions

lainsce:
Possible fix for #667, makes the symbolic versions look like the port, in the orientation used in the color versions.

For comparison:

| Color | Symbolic |
| --- | --- |
| network-wired | network-wired-symbolic |

Screenshot in action:

screenshot from 2018-09-24 12-51-02


waldyrious:
I'm afraid this makes the icon less intuitive than the current one. If anything, I'd support converging to the same representation by changing the colored versions, rather than the opposite.

lainsce:
@waldyrious Let's wait to see what the elementary UX team has to say.

danrabbit:
I agree that this is a step backward from the current situation. Something to consider as well is that it looks like tethered USB connections also use this namespace. So it's probably not accurate to always use the RJ45 metaphor at all.

lainsce:
Tethered USB connections should use another colored icon/symbolic icon possibly? As it stands, they use the RJ45 colored/symbolic icon, not their own, which might make it strange on that side, actually.

lainsce:
This is a step backward, we discussed a better plan on Slack. Closing.

PR #670, opened 2018-09-24
New colored icons for Ethernet devices without hardware metaphor

lainsce:
Removing the hardware RJ45 port metaphor for ethernet devices and using a generic Wired metaphor instead.

It looks like this:
screenshot from 2018-09-24 15 28 05


lainsce:
Micah in Slack suggested the orange to bring som warm colors to the Network plug, and I went with it.

waldyrious:
I like the shape, but I'm not sure about the color. I typically associate traffic light-like colors (red, orange/yellow, green) to connection status in network contexts. I understand the appeal of using a warm color, but I fear that the orange may unconsciously confer a slowed-down feeling to the icon.

I suspect the "Network Purple" (one of the "Grape" shades in the [color palette](https://elementary.io/docs/human-interface-guidelines#color), I suppose?) might work better here after all.

lainsce:
Yeah, Network Purple is the Grape in the palette.

I've grown fond of the orange, it doesn't look like the connection is failing due to the online dot beside it, which by the way is even more visible with the orange than the purple.

micahilbery:
Common traffic light colors are usually red(stop), yellow(yield), and green(go) and that is why I didn't suggest red or yellow. I thought orange would bring in some warmth with out too many negative connotations. I realize that orange can be used to convey warning but in this context there is a status indicator with it that should mitigate too much confusion.

waldyrious:
Fair enough. It wasn't a strong objection on my part, just a slight uneasiness for a possible misinterpretation. If you guys aren't concerned, it probably means my concerns were unfounded :) so FWIW, this gets a :+1: from me!

lainsce:
Speaking of which I'm actually using the Music square here :blush:

lainsce:
Is this not an option anymore due to the other PR?

cassidyjames:
Superseded by #732

PR #671, opened 2018-09-24
Create specialized VPN icons

danrabbit:
Instead of recycling the lock, create a new icon based on the new security icons

## OLD:
screenshot from 2018-09-24 10 08 37 2x

## NEW:
screenshot from 2018-09-24 10 05 37 2x


lainsce:
Why not use the "Network Purple" here? It'd differentiate this icon from the generic security lock icon.

PR #672, opened 2018-09-25
New Wi-Fi icons for the network plug

lainsce:
Remade the network Wi-Fi icons to make them cohesive with other network icons in the plug, as well as updating it to look more in-line with the newly made icons & palette.

In action:
screenshot from 2018-09-25 09 44 19


danrabbit:
I’m -1 here. I think adding another shape is a step backwards

micahilbery:
Yeah I'm -1 here as well. I'm not really in favor of loosing another unique shape for to another circle or square.

Issue #440, opened 2018-02-27
preferences-desktop-workspaces


danrabbit:
Or something like this name. Right now we're abusing preferences-desktop-wallpaper for the keyboard shortcuts navigation.

Astro-City:
Hi.  

This would be great.  

I see some people that don't like the fullscreen=new workspace or would want a static/dynamic toggle.

PR #676, opened 2018-09-26
Make new icons for Workspace preferences

lainsce:
Make a new iconset for Workspaces preferences.
Fix for issue #440.

Looks like this in action:
screenshot from 2018-09-26 22 51 25


waldyrious:
I think it would be nice to have a bit more contrast in the small grey squares in the bottom.

Issue #539, opened 2018-05-13
missing actions/mail-mark-junk


danirabbit:
and mail-mark-notjunk

from the fd.o spec:

"The icon for the mark as junk action of an electronic mail application."

cassidyjames:
We should decide what we want as the metaphor for this. Our symbolic version is a stop sign with an exclamation point, which is what at least Gmail also uses. Do we want to make a color version of that?

I think we want to avoid a trash can as it could be confused with actual trash. GNOME uses a wadded up piece of paper. Fastmail uses a shield, which is interesting (I think web services sometimes do that as they're using it to represent their spam filter, shielding you from junk).

cassidyjames:
If it makes sense, we could also do a mail letter badged with the stop sign (mark-junk) or a check mark (mark-not-junk).

lainsce:
Hailing from my PR, I've noticed most OSes use either a trashcan with a mail envelope sticking out, or a trash recycling bag with the same envelope sticking out.

danrabbit:
I think since we're in the context of mail here,  we don't need to include an envelope as part of the metaphor. In Mail, these actions are actually behind the "Flag" menu and the junk folder is represented by the flag icon

lainsce:
Oh, then it's even simpler. A simple "!" in a red or grey octagon depending on marking or not would suffice then.

PR #677, opened 2018-09-26
Made new icons for marking/not marking mail as junk

lainsce:
Made a new iconset for these actions.
Fix for issue #539.

Looks like this:
mark-mail-junk =
screenshot from 2018-09-27 23 11 45

mark-mail-notjunk =
screenshot from 2018-09-27 23 11 56


cassidyjames:
@lainsce junk is distinct from trash, so I don't think we want to confuse those two.

lainsce:
All the iconsets I've seen online for the junk metaphor use a trashcan or a trash bag with a mail envelope sticking out. :thinking:

cassidyjames:
@lainsce see #539 for icon discussion, as I want to get a consensus on the metaphor before we get too far. :)

lainsce:
Following the discussion, it seems we've arrived in a common ground metaphor, the Octagon.

micahilbery:
I think the exclamation mark is a little inconsistent I would grab the one from `status` / `dialog-error`. Also I think the stroke color on the red is a little dark I would look at the stroke and shadow colors from that icon too as it's also red.

lainsce:
The exclamation mark did indeed come from `status/dialog-error` :grimacing: 
Also, I'm just using the darkest red in the official pallette as is common to use the darkest one.

micahilbery:
I'm talking about the non-symbolic version.

dialog-error

lainsce:
Oh, that one. I'll use it then :smile:

danrabbit:
Please be mindful of the grid

screenshot from 2018-09-27 09 50 39 2x

lainsce:
Now it should look amazin'.

danrabbit:
Some of your nodes are still off:

screenshot from 2018-09-27 16 09 32 2x

Please make sure you're using the same 4 stop gradient for highlights instead of a new 2 stop gradient:

screenshot from 2018-09-27 16 10 42 2x

Also please keep the same shadows from dialog-error instead of inventing a new shadow. Especially since these shadows are a darker and more saturated red and yours are black:

screenshot from 2018-09-27 16 11 59 2x

lainsce:
It seems the war isn't over yet. I'll fight until the end and hopefully win!

danrabbit:
Getting closer! It looks like none of these icons are equilateral

lainsce:
It's a gamble, grid-respecting, or equilaterality :grimacing:

danrabbit:
I think equilateral is probably more important than the diagonals behind at grid corners. Diagonal lines have to be antialised anyways.

Something is really funky with the placement of the exclamation point on your 48px icon. Compare to the right side which is equilateral and the exclamation sits in the same position as dialog-error:

screenshot from 2018-09-27 17 54 26 2x

lainsce:
I'll look into it.

lainsce:
Finished redrawing all of them to be equilateral while also minding the grid.

lainsce:
I'll look into it when I can, but I remember them being equilateral.

tintou:
May I ask why use an octagon?
The only octagon that I know is the stop road sign :man_shrugging:

lainsce:
Octagon in this case is an universal icon in an email setting, just like using a paper airplane for "Send".

lainsce:
:smile:

PR #683, opened 2018-09-29
Add mail-mark-unread icons

### LINKED ISSUE #678: mail-mark-read
Looks like we're falling back on the GNOME one, and there's no symbolic. I think an envelope is probably fine as a base. We could do an envelope badged with a red dot, like we treat the unread notifications icon (and similar to dock badges).

lainsce:
Adds icons for the mail-mark-unread iconset.
Fixes #678.

Looks like this:
screenshot from 2018-09-29 01 16 39


lainsce:
I want a review of this following the review on #684.

fabiozaramella (approved):
Same as in #684  👍🏻

PR #684, opened 2018-09-29
Add mail-mark-read iconset

### LINKED ISSUE #679: mail-mark-unread
Looks like we're falling back on the GNOME one, and there's no symbolic. I think an envelope is probably fine as a base. 

We could do the open envelope with a document sticking out.

lainsce:
Adds icons for the mail-mark-read iconset.
Fixes #679.

Looks like this:
screenshot from 2018-09-29 04 29 56


lainsce:
The current fallback iconset does the same, it has the unread have the "read" icon, and read have the "unread" icon, check it out on Lookbook. :confused:

danrabbit:
I think you're mistaken. Here's `mail-mark-unread` in Adwaita:

mail-mark-unread

lainsce:
When this PR was made, Lookbook had them both reversed then or the fallback iconset had them reversed.

fabiozaramella (approved):
Looks good to me ❤ 
I hope someone from @elementary/ux team will find time to approve it and #683 as well.

PR #685, opened 2018-09-30
Made a colored call-stop & call-end icon

### LINKED ISSUE #534: Missing actions/call-stop
From the fd.o spec:
"The icon used for stopping a current call. Should be similar to the standard cellular call hangup icon, a red handset with ear and mouth pieces facing downward."

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58294847-missing-actions-call-stop?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lainsce:
Made an icon for the call-stop spec.
Fixes #534.
**NOTE:** neither call-stop or call-start have icons above 16px size.

Looks like this:
screenshot from 2018-09-29 22 20 54

### DIFF
--- a/actions/16/call-end.svg
+++ b/actions/16/call-end.svg
@@ -1,0 +1,1 @@
+call-stop.svg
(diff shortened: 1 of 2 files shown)


danrabbit:
Is there some reason for the symlink `call-end`?

lainsce:
It has a symbolic icon which is also a symlink to call-stop-symbolic, but no colored icon, thus, I thought it should be the same in color icons.

PR #688, opened 2018-09-30
Make new colored icons for format-text-direction-ltr/rtl

### LINKED ISSUE #538: Missing actions/format-text-direction-ltr
and -rtl

From the fd.o spec:

"The icon for the left-to-right text formatting action."

lainsce:
Made new colored icons for these actions.
Fixes #538.

Looks like this:
LTR = 
ltr

RTL =
rtl


danrabbit:
Looking at some other icons that have been done:


I think something like this is interesting because it uses the lines as arrows


034_098_format_text_direction_left-128
Something in this direction seems like it probably works really well at small sizes

My guess is that this icon would more likely be used in the 16px symbolic variant, so I think it's probably important to make sure the metaphor works really well in a symbolic format

lainsce:
Could always expand and someone does a symbolic variant of this PR.

lainsce:
Expanded and made symbolics as well, using the ideas from other format icons.

danrabbit:
I'm going to close this since the metaphor here is less clear than what is already in

PR #731, opened 2018-11-29
Redesign Phone icons

danrabbit:
because phones look a bit different these days

### DIFF
--- a/devices/128/phone.svg
+++ b/devices/128/phone.svg
@@ -25,14 +25,14 @@
      inkscape:pageopacity="0"
      inkscape:pageshadow="2"
      inkscape:window-width="3200"
-     inkscape:window-height="1674"
+     inkscape:window-height="1686"
      id="namedview64"
      showgrid="false"
-     inkscape:zoom="2"
-     inkscape:cx="95.031886"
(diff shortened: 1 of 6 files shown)


cassidyjames:
Another possibility would be the back-side of a generic smartphone, since most of those have at least one camera (often in the top-left) and could give you some more details and color to work with like a flash, antennae lines, etc.

NikoKrause:
No notch

danrabbit:
:man_shrugging: I feel like it's not hard to push another rev when phones stop looking like that but the current flagship generation is all notches. Either way, removed

waldyrious:
The front bezels haven't completely receded yet, notch nonwithstanding (they have become mostly black, indeed). That, plus the fact that an almost completely featureless black rectangle may not be very illustrative, how about lighting up the screen? For example, these are the top results from a google image search for "smartphone icon" I just performed:

screenshot_2018-11-30 smartphone icon - google search

I see several potentially interesting options here. We could go with (1) a white screen, as most of these do; (2) some abstract interface; (3) a color gradient; (4) a solid color. IMO any of these would be more obvious than a black rectangle, and would still represent the current look of smartphones.

PR #732, opened 2018-11-30
Redesign wired networking icons

### LINKED ISSUE #667: Wired internet icons don't match between colored and symbolic

danrabbit:
Fixes #667 
Supersedes #670 

Uses a simpler wired networking metaphor that scales better to different sizes and doesn't imply any specific hardware setup. Also uses the consistent slashy metaphor

screenshot from 2018-11-29 15 54 28 2x


cassidyjames:
I don't care enough to hold this up. :stuck_out_tongue:

cassidyjames:
In context in the panel:

Screenshot from 2019-08-04 22 30 00@2x

KomalleanaByzantis:
@danrabbit @cassidyjames Was this the total sum of the discussion that was had before it was decided that elementary OS is to deviate from the ISO standard? Is elementary OS to deviate from a standard merely due to a dislike of that standard? The standards exist so that companies large and small cannot simply decide on their own to set their own unofficial standard and hope or expect others to automatically follow suit. I dearly wish there were not such a disregardful attitude to the importance of adherence to official standards. I am much in favour of the solution that Blaede suggested, such that elementary OS would return to being compliant with the standard.

If standards-compliance is not a priority, then where does the line go for what is sufficiently “inconvenient” to be discarded? What visual representation defined by ISO will be the next to undergo a change because someone simply did not like the standard? What if everyone in the industries of computers, phones, or other appliances did that? Such a scenario would be very unfortunate and unintuitive to consumers. Surely elementary OS should not pretend to have the special right to be different, nor should Apple or Google. Competition in symbols is not favourable; instead, artistic licence in how a symbol is rendered is to be taken so that it may look properly scaled.

I am aware that this is an old issue. Is the issue of this icon “settled” and thereby irreversible in the eyes of the developers? Is this up for debate? Would opening a new issue make any difference?

cassidyjames:
@KomalleanaByzantis there was more discussion offline than what is listed here (and I should have put a summary of the discussion here, apologies for not doing so). Essentially, we discovered that macOS, iOS, iPadOS, Android, and Chrome OS all use this symbol to represent a wired network connection. The ISO standard was not explicitly designed for display in UI for representing wired networks, but was designed to represent any computer network and/or the physical connections.

Sometimes our job as designers of an OS is to decide when to follow and apply certain standards, when to do our own thing that we believe is better, and when to fall in line with established conventions. In this case, we're aligning ourselves with some of the most popular platforms' conventions while also solving a problem with an overly-complex icon that did not scale well. There's also nothing about the cited ISO standard that says it is for display in computer operating systems, and it seems that no other major operating system had interpreted the standard in that way.

KomalleanaByzantis:
@cassidyjames It seems strange to me that four mobile devices and one web browser-based operating system are “some of the most popular platforms” that have led to this outcome, given that elementary OS is a desktop operating system. What sense does it make to follow the non-standard conventions that Apple and Google have decided to use on these platforms when those platforms fundamentally differ from elementary OS in how they function and are used? Why not simply redesign the way the old icons appeared so that they would look pleasing in all sizes, rather than abandoning them altogether in favour of an icon that is alien to many and non-standard? Would you say that there is a bias within the design team of elementary, Inc. towards Apple and/or Google products, designs, and conventions?

Finally, you never answered these questions: Is the issue of this new icon “settled” and thereby irreversible in the eyes of the developers? Is this up for debate? Would opening a new issue make any difference?

I do not mean to cause needless discussion if my words as a user of your operating system will make no difference whatsoever; if this is the case, please let me know.

cassidyjames:
>It seems strange to me that four mobile devices and one web browser-based operating system are “some of the most popular platforms” that have led to this outcome, given that elementary OS is a desktop operating system. 

macOS and Chrome OS are desktop operating systems that run on laptops and desktops, just like elementary OS. Android is used on a wide variety of hardware and supports desktop computing, including Ethernet out of the box. Android is the most popular OS in the world. We looked at Windows (the most popular desktop OS) as well, but their metaphor was even messier representing a physical port with a display overlaid, so we didn't think it was a convention worth following.

>What sense does it make to follow the non-standard conventions that Apple and Google have decided to use on these platforms when those platforms fundamentally differ from elementary OS in how they function and are used?

I disagree with the premise that those platforms fundamentally differ from elementary OS.

>Why not simply redesign the way the old icons appeared so that they would look pleasing in all sizes, rather than abandoning them altogether in favour of an icon that is alien to many and non-standard?

First, we tried that and it was not successful. Due to the representation of the symbol, it's impossible to satisfactorily pixel-align it to prevent blurriness. If we invented our own pixel-aligned version it would need to be substantially different, and at that point it would just be inventing our own symbol anyway. We might as well use a conventional icon used by other platforms instead.

 >Would you say that there is a bias within the design team of elementary, Inc. towards Apple and/or Google products, designs, and conventions?

No. I would say there is a "bias" towards looking at already-solved problems elsewhere and that's part of what makes elementary OS what it is; instead of inventing everything new from scratch, we often look to other platforms and combine some of the best concepts into one cohesive vision. Just like app menus (inspired by Google Chrome at a time when desktop apps all used menu bars), overlay scrollbars (inspired by mobile OSes at a time when desktop OSes still used chunky space-wasting scrollbars), etc. We are familiar with other platforms like Android, iOS, macOS, Windows, Chrome OS, etc. and are not afraid to borrow ideas from any of them when it makes sense and would make elementary OS a better product.

>Finally, you never answered these questions: Is the issue of this new icon “settled” and thereby irreversible in the eyes of the developers? Is this up for debate? Would opening a new issue make any difference?

Nothing is permanent; the nature of elementary OS is that it constantly evolves and improves. However, I highly doubt that this specific issue will be reversed because there has not been a compelling argument that overrules the use of a widely-adopted convention that also solves the scaling issue.

PR #737, opened 2018-12-05
Install distributor-logo icons to hicolor directory

### LINKED ISSUE #377: Split out elementary logos into their own repository/package
>If one doesn't take care to fix this downstream, it's entirely possible to unintentionally use the elementary logo in places they shouldn't be used.
>
>To avoid this, the elementary logos should be removed from this repository and put into their own overlay package that elementary OS uses. That way, when the theme is packaged, the logo doesn't get included unless someone deliberately adds it.

–@Conan-Kudo from https://github.com/elementary/stylesheet/issues/247

codygarver:
Fixes #377


codygarver:
I tried different approaches in different packages to solve this and decide it was better to let them live in this repo since they are icons but also to split them to a different package, which will happen in the packaging branch should this get merged. That will allow potential distributors/OEMs/whatever to easily exclude these from elementary OS if they want, without forking this repo to disable them.

I tried to get `install_subdirectory` to work with an array to make this fewer lines but it doesn't seem possible. Feedback requested from @tintou 

Also the icon directory names are different than elementary theme to match hicolors and make installation a simple copy and paste.

codygarver:
Fixes #377

tintou:
I don't get why that's even necessary when distributor *have* to patch this repository anyway

codygarver:
@tintou in what ways is it necessary for downstreams to patch this icon set? The volumeicon is optional, that's all I can think would cause a problem if it weren't.

Also note in the bug #377 there's a case for splitting this icon out to an optional package, so that a hardware vendor can slot in their own.

tintou:
This PR doesn't move the icon in a separate package so It doesn't solve anything, people can just explicitly ignore the icon in the packaging if they don't want it.

codygarver:
@tintou 

One of the use cases, a hardware vendor shipping elementary OS, cannot explicitly ignore the icon, because it's globbed into the main icon theme package.

See PR #741

tintou:
Since all the distribution are going to use the LOGO field, there is no need for the distributor-logo. We should provide LOGO=elementary and use this field only.

codygarver:
Yes, I agree, that makes sense and is a good middleground for both of our concerns.

PR #752, opened 2019-01-10
Add 16px night-light icons

ochosi:
Not sure if you even care about 16px versions of the symbolic status icons you have in 24 now. However I wanted/needed one for night-light (I also want to use the icon in the xfce panel, which can have 16px symbolic icons) so I created it.
If you generally want this icon, I can also create the missing files night-light-disabled-[10..90]-symbolic.svg


danrabbit:
Thanks! I don't think the animation steps are worth the time unless you're using them

ochosi:
Yeah, I guess we're not using animation for now. Thanks for merging!

PR #778, opened 2019-02-10
Add missing application-x-generic

danrabbit:
It looks like GLib.FileInfo returns this icon as a fallback for mimetype icons, so we need to have it


waldyrious:
Wouldn't [application-default-icon](https://github.com/elementary/icons/blob/master/apps/32/application-default-icon.svg) be a better target for `application-x-generic`?

PR #784, opened 2019-02-20
Add location status icons

danrabbit:


waldyrious:
Isn't this shape more associated with routing / directions (or even with a generic "send" action) than with static location? I would have gone with the pin icon.

danrabbit:
@waldyrious this is the icon we use for location services. See System Settings → Security & Privacy

waldyrious:
I understand, but I believe my point stands regardless. If I understand correctly, changes in icon shapes are acceptable if justified, as was the case with #670 / #732. Or is this different?

PR #803, opened 2019-06-23
preferences-system-notifications as a bell

danirabbit:
I'm not sure I'm super happy with the colors, but I feel like the metaphor is a step forward

Drafting for feedback

Context:

Screenshot from 2019-06-22 18 50 35@2x


cassidyjames:
Yeah I'm :+1: for the improved metaphor. I do think the colors and possibly perspective could be tweaked. What if it was ringing (angled sideways) like the symbolic one?

waldyrious:
@danirabbit since #1234 was merged (What a nice PR number, btw! Congrats, @newhoa :grin:), shouldn't this one be closed?

danirabbit:
I'd like to still iterate on it a bit so you I'll leave this open for now

PR #812, opened 2019-08-18
Make the file icon less bland and more vibrant

Fatih20:
The original design wasn't bad. I'm just thinking that the color could be... better. It just has that taste of bitter like when you're drinking cough syrup. Subjective indeed, any thought?

### DIFF
--- a/apps/128/system-file-manager.svg
+++ b/apps/128/system-file-manager.svg
@@ -6,12 +6,47 @@
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    version="1.1"
    width="128"
    height="128"
-   id="svg4000">
+   id="svg4000"
+   sodipodi:docname="system-file-manager.svg"
+   inkscape:version="0.92.4 5da689c313, 2019-01-14">
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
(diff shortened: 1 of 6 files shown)


cassidyjames:
The current icon is intentionally silver, as it's a system utility.

Fatih20:
What about system settings? The dominant color when you look at it is blue.

cassidyjames:
It has blue because that's the color of switches. :smile: I guess the inset home icon could be blue to sort of evoke that, but I don't really think it's necessary. @elementary/ux thoughts?

Fatih20:
I think that would be great. It's fine if you want it to be silver but right now it looks more like a decayed paper which shouldn't be how we look at opening files.

waldyrious:
I'd expect it to use the same color as the folder icons in Files, tbh.

Fatih20:
What would you say if we just recolour all of the folder? Because the file manager icon looks like a paper that is about to decay, the actual file icon look like a paper that has already decayed for 2 years at least.
folder

danrabbit:
Now that the standard folder is Manila I wouldn’t be opposed to the Files icon using palette blue (or another palette color). Could be fun :)

Fatih20:
So, do you want me to make a recolor based on the sample I've sent?

danrabbit:
@Fatih20 I've proposed #813 for discussion which uses our palette blue color

Fatih20:
I'll start recoloring all of the folder based on that then.

Fatih20:
> @Fatih20 I've proposed #813 for discussion which uses our palette blue color

But do you think that the color of folders icons need to be blue as well like the file manager icon you proposed? If so, I'll start recolor them and post the result for discussion.

cassidyjames:
@Fatih20 I suggest you look at https://github.com/elementary/icons/pull/634 ;)

We used to have blue folders, but moved to manila to better match real-world folders, to make the overall desktop less overwhelmingly blue, and to support potential per-content colors.

PR #813, opened 2019-08-21
Change file manager to blue

danirabbit:
Supersedes #812


micahilbery:
What it looks like in context:
Screenshot from 2019-08-21 14 31 35

micahilbery:
I'm fine with the color change and the larger house on the 32px icon is definitely a nice improvement :+1:

jeremypw:
I would change the imprint to something more generic to indicate "navigating" or "browsing" or "searching", rather than a particular location.

Fatih20:
I agree. Look how the files manager feels out of place in this screenshot.
Screenshot from 2019-08-30 16 58 51

Fatih20:
> I would change the imprint to something more generic to indicate "navigating" or "browsing" or "searching", rather than a particular location.

What do you think of this design? It's supposed to follow the principle used like the appcenter icon where it looks like an actual store. Same thing with videos icon, it looks like an actual video player.
file-manager-eOS-m

Fatih20:
file-manager-eOS-s
Or this maybe?

Fatih20:
file-manager-eOS-mc
What do you guys think?

danirabbit:
Closing since we got in a new files icon and it is blue :)

waldyrious:
> Closing since we got in a new files icon and it is blue :)

For reference: #1147

PR #834, opened 2019-11-01
appdata: remove U+200B (zero-width-space) characters

### LINKED ISSUE #831: Zero-Width-Space Unicode characters (U+200B) in appdata breaking libappstream-glib
I have no idea why this keeps happening in new appdata files (I assume it's a borked template somewhere?). But the U+200B characters (that are visible in the file e.g. when editing it in vim) are breaking the GLib XML parser and make the appdata file invalid when trying to validate it with `appstream-util validate-relax` (and presumably this would also break gnome-software and other programs handling appdata files via libappstream-glib).

decathorpe:
Fixes #831


cassidyjames:
Whoops, didn't see you'd PR'd this. Fixed in master.

PR #851, opened 2019-12-07
Add cursor sources

### LINKED ISSUE #843: Build cursors from source
Sam found sources here: https://github.com/kaesetoast/elementary-cursors

We should use our fancy build system to generate the cursors and replace what we have now with the actual editable sources 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/84259397-build-cursors-from-source?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #843


tintou:
This also needs a change to the packaging to add the required utilities.
Can you replace:
```meson
inkscape, '-z',
'-w', size, '-h', size,
'-e', join_paths (meson.build_root(), 'cursors', size + 'x' + size, cursor + '.png'),
join_paths ('24', cursor + '.svg')
```
To:
```meson
inkscape, '--without-gui',
'--export-width=@0@'.format(size), --export-height=@0@'.format(size),
'--export-png=@0@.png'.format(join_paths (meson.build_root(), 'cursors', size + 'x' + size, cursor)),
join_paths ('24', cursor + '.svg')
```
That way it is faster to understand the command later (especially with the `-z` meaning `--without-gui`)

PR #873, opened 2020-02-18
Cursors/meson: Use Flatpak'd Inkscape

cassidyjames:
Not 100% this is the best way, but it seems to work.


cassidyjames:
I'm also not 100% what the failure case is for not having Inkscape installed via flatpak, or how to test for that.

danrabbit:
You'll need to update the CI environment to include the inkscape flatpak

Also, we'll need to update the debian packaging to pull in the inkscape flatpak somehow

danrabbit:
I'm going to close this since it's not passing CI and there hasn't been any more progress

PR #874, opened 2020-02-18
Add 32px system-os-installer

danrabbit:
I have no idea what this was for, but figured I'd make a draft request in case anyone remembers


danrabbit:
Gonna close since we never figured out what it was for

PR #875, opened 2020-02-18
play around with a more organic check shape

danrabbit:


danrabbit:
Closing because it never went anywhere

PR #888, opened 2020-05-02
Use librsvg instead of inkscape

### LINKED ISSUE #886: make inkscape depedency optional
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
since #851 inkscape became a required depedency, making the flatpak-platform fails to build using more recent releases. since it's only used to make the cursors make the depedency optional and only build the cursors if inkscape is installed.

### To Reproduce
1. Change the Flatpak manifest in [elementay/flatpak-platform](https://github.com/elementary/flatpak-platform) to build the latest release
2. try to build
3. flatpak-builder fails because it's don't have inkscape installed

### Expected behavior
build icons skipping the cursors folder (the unique that depends on inkscape)

### Please check what applies:
- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

MrMarukesu:
Fix #886.

### DIFF
--- a/cursors/meson.build
+++ b/cursors/meson.build
@@ -1,5 +1,5 @@
 cursor_dir = join_paths(icon_dir, 'cursors')
-inkscape = find_program('inkscape')
+rsvg = find_program('rsvg-convert')
 xcursorgen = find_program('xcursorgen')
 
 cursors = [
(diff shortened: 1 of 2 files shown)


cursors/meson.build
```diff
@@ -100,9 +100,9 @@ foreach size: sizes
             message('rendering: ' + size + 'x' + size + '/' + cursor + '.png')
 
             run_command(
-                inkscape, '--without-gui',
-                '--export-width=@0@'.format(size), '--export-height=@0@'.format(size),
-                '--export-png=@0@.png'.format(join_paths (meson.build_root(), 'cursors', size + 'x' + size, cursor)),
+                rsvg,
```
alatiera, line 103: `rsvg-convert`defaults to .png when no format is specified for output but I'd pass `--format png` into it just in case.

Another thing that might be problematic, is that it defaults to `--keep-aspect-ratio FALSE`. I don't know what inkscape defaults to, so somebody should double check that probably.

MrMarukesu: inkscape doesn't keep the aspect when exporting from command line, neither have a option for that.

alatiera:
Looks good :+1:

PR #900, opened 2020-05-21
Add office-database-new icons

Alecaddd:
I'd like to propose the addition of the `office-database-new` mime icons.
This would be useful for SQL application, like Sequeler and Peeq, for toolbar buttons to create a new database.

Preview:
office-database-new


micahilbery:
I don't think these would go in with mimes. This feels like it would be more appropriate to be in actions.

Alecaddd:
Sounds good, I'll move them there.

Alecaddd:
Updated!

PR #907, opened 2020-07-01
Add Inkscape's Text Alignment Icons

### LINKED ISSUE #762: Text alignment actions
- [ ] align-horizontal-baseline
- [ ] align-horizontal-baseline-symbolic
- [ ] align-vertical-baseline
- [ ] align-vertical-baseline-symbolic

see http://wiki.inkscape.org/wiki/index.php/Themable_icons

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68955858-text-alignment-actions?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Fatih20:
Fixes #762

This is how the icons looked like under elementaryOS light theme
image
This is how the icons looked like under elementaryOS dark theme
image
This is how the symbolic icons looked like under elementaryOS light theme
image
This is how the symbolic icons looked like under elementaryOS dark theme
image

In the file there are a lot of other alignment icons in a separate layers. I was using those to line the details up since they are located side-by-side with the text alignment button in Inkscape. I will clean them up once this is finalized and approved.


Fatih20:
> I think for baseline alignment especially we need to use a letter that has a descender. This is probably why the original icon uses "y". Otherwise it's not exactly as clear that the alignment is to the baseline and not to the absolute bottom of the path

You're right, I thought about that too. I also wonder what is the font that you use in `draw-text`, seems to be an ideal font for making icons.

Fatih20:
Alright @danrabbit.

I've done basic research about [baseline](https://en.wikipedia.org/wiki/Baseline_(typography)). These letters are the most common to have descenders, g j p q y. I decided to not pick g, p, and q because they have "hole" in it that would make them hard to recognize since the icons in question have small sizes. Especially since elementary icons style has outline, which would fill up said hole and worsen its recognisability. Meanwhile j would be easily mistaken for a line, the dot at the top also aren't easily perceived. These are in contrast to the simplicity of y. The non-descender letter I chose is z, because it's simple and more coherent than the a in the Inkscape's built-in icons since z comes after y. I looked up font style with the biggest descenders of y and open-source or free license at dafont and found [Highway Gothic](https://www.dafont.com/highway-gothic.font?l[]=10&l[]=1&text=yz). After some experimentation it looks pretty good. Here's a screenshot of Inkscape using it and the icon itself.
image
I made the text gray instead of black like both Tango and Inkscape did, because I wanted it to look like a regular object (in case you think this is a bad idea because text needs to be distinct for whatever reason, I could change it). 

The problem is, because I had to include the descenders, the letters are taller than they have to be otherwise. This makes it hard for designing the `align-horizontal-baseline`. As you can see from this screenshot, the letters would intersect each other.
image
 Furthermore, because the nature of y and z I've picked, it doesn't show the horizontal version of a descenders, making it look like just `align-horizontal-left`, repeating your criticism of my previous proposal. My solution to this is to pick a different letter (or even font style) that could properly emphasize this. However, both Tango and Inkscape's built-in icons doesn't do this, they use the same letter. So, before doing this, I'd like to ask if this is a good idea or should I just use the y and z for the horizontal baseline icons?

danrabbit:
Messed around a bit and it seems like ubuntu bold italic is pretty good here as far as being a fairly uniform shape and a y that sticks out pretty far:
baselines

I think "a" works better than "z" here since the round shape pulls it away from the guide line

[baselines.svg.tar.gz](https://github.com/elementary/icons/files/4898933/baselines.svg.tar.gz)

Fatih20:
This is how the icons looked like under elementaryOS light theme
image
This is how the icons looked like under elementaryOS dark theme
image
This is how the symbolic icons looked like under elementaryOS light theme
image
This is how the symbolic icons looked like under elementaryOS dark theme
image
As you know, the symbolic icons only have one color, to avoid making the y unrecognizable because it's muddled with the baseline, I used negative space to create an illusion of a border.

In the file there are a lot of other alignment icons in a separate layers. I was using those to line the details up since they are located side-by-side with the text alignment button in Inkscape. I will clean them up once this is finalized and approved.

Fatih20:
I come across the problem with the color though. As you know, we want to emphasize how the descenders intersect the baseline. I run the test and the orange-gray combination that I used failed the all of the accessibility test. As such, it would be borderline unseen that the descenders of the letter y is intersecting the baseline, even for me who have a normal vision. I think we'd need to pick another color.

Fatih20:
This is how the icons looked like under elementaryOS light theme
image
This is how the icons looked like under elementaryOS dark theme
image

Symbolic icons is unchanged from my previous screenshot.

In the file there are a lot of other alignment icons in a separate layers. I was using those to line the details up since they are located side-by-side with the text alignment button in Inkscape. I will clean them up once this is finalized and approved.

Fatih20:
Before committing, here's how it looked like under both dark and light theme.
image
image
This is from `draw-text`, I haven't changed the color of the other icons so you can compare them.

danrabbit:
@Fatih20 Please make sure to use whole pixel width strokes for the highlights. I'll review again once you've pushed

Fatih20:
> @Fatih20 Please make sure to use whole pixel width strokes for the highlights. I'll review again once you've pushed

What do you mean Dan?

danrabbit:
You can see in your screenshots that the highlights on the letter forms are not 1px wide, but only a portion of a pixel

Fatih20:
Here's how it looked like now.
image
image

danrabbit:
I'll be happy to review again once you've committed your changes

danrabbit:
Finished in https://github.com/elementary/icons/pull/968

Closing

Issue #713, opened 2018-11-02
Color picker


danrabbit:
an color action icon named `color-picker`.

It looks like we currently have a symbolic icon `color-select-symbolic`. It may be that apps are using both of these icon names so we should probably symlink

Fatih20:
Can you elaborate? I wanted to work on the `color-picker`.

Fatih20:
Can you close this issue? This is done.

PR #914, opened 2020-07-11
Add Inkscape's Color Picker

Fatih20:
This is how the icons look under elementary's light theme
image
This is how the icons look under elementary's dark theme
image

There is some things that is still in the file that I use in the design process of these icons. I'll be cleaning them up once these are finalized and approved.


Fatih20:
I also wanted to redesign `color-select-symbolic` and make a `color-picker-symbolic` symlink for it. Is that okay?

Fatih20:
This is how the icons look under elementary's light theme
image
This is how the icons look under elementary's dark theme
image

There is some things that is still in the file that I use in the design process of these icons. I'll be cleaning them up once these are finalized and approved.

danrabbit:
Looks like in this latest revision your highlights are all missing as well

Fatih20:
Can you elaborate? I don't understand your comment. @danrabbit 

> It looks like you have the fill line on a half pixel

What do you mean?

danrabbit:
Sure, as you can see here, the edges of the stroke are not grid aligned and instead sit a half pixel off:

Screenshot from 2020-07-14 15 26 22@2x

Fatih20:
I've lowered by it by half a pixel and fixed the disappearing liquid's highlight.

danrabbit:
Looks good, will merge after cleanup! Thanks :)

Fatih20:
Done.

Fatih20:
Can you close #713 ?

PR #467, opened 2018-03-30
Add a bunch of graphics tool icons

danrabbit:
Based on Tango icons for Inkscape

- [x] color fill
- [x] color-gradient
- [x] color-gradient-mesh
- [ ] color-picker
- [ ] draw-calligraphic
- [x] draw-cuboid
- [x] draw-ellipse
- [x] draw-eraser
- [x] draw-freehand
- [x] draw-path
- [x] draw-polygon-star
- [x] draw-rectangle
- [x] draw-spiral
- [x] draw-text
- [x] tool-measure
- [x] tool-node-editor
- [x] tool-pointer
- [ ] tool-spray
- [ ] tool-tweak
- [ ] zoom

PR #915, opened 2020-07-11
Add Inkscape's Zoom

Fatih20:
Continues upon #467 

This is how the icons look under elementary's light theme
image
This is how the icons look under elementary's dark theme
image

### DIFF
--- a/actions/24/zoom.svg
+++ b/actions/24/zoom.svg
@@ -1,0 +1,488 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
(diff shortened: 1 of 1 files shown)


cassidyjames:
Is there a reason we're not symlinking to an existing zoom icon? I am also concerned about the similarities between zoom and search here.

Fatih20:
> Is there a reason we're not symlinking to an existing zoom icon? I am also concerned about the similarities between zoom and search here.
Is there any other zoom icon other than the one with white square? Sorry if I'm not being thorough.
I was basing my design in the context of Inkscape. Which, because of the icon placement probably wouldn't be mistaken for search. However, I think zoom is already synonymous with magnifying glasses (unfortunately, so does search which created the problem you've pointed out), that making the design something else would be weird, no?

jeremypw:
@cassidyjames This review seems to have stalled - should it be closed?

kurisubrooks:
I think to improve the clarity of the meaning while also visually separating it from a "search" icon, we could re-use the [edit-find](https://github.com/elementary/icons/blob/main/actions/24/edit-find.svg) (magnifying glass) icon with a [plus](https://github.com/elementary/icons/blob/main/actions/24/list-add.svg)/[minus](https://github.com/elementary/icons/blob/main/actions/24/list-remove.svg) symbol to represent the action of zooming.

Alternatively for improved skeuomorphic representation/meaning, we could layer some sort of generic "photo"/image representation underneath the magnifying glass icon, with the inner-glass circle being a more zoomed in version of that, but this style may have issues with contrast and complexity at low resolutions.

PR #917, opened 2020-07-13
Refresh pane-show and pane-hide symbolics

lainsce:
Made them follow most newer symbolics, by dropping the 2px thick pane in favor of a 1px thick pane symbol. (RTL and LTR included)

Looks like (in a setting comparing with other symbolics):
`pane-show-symbolic`
pane-show-preview

`pane-hide-symbolic`
pane-hide-preview


danrabbit (changes requested):
Since we have new arrows, these should be updated to match the new arrow shape

lainsce:
Done.

danrabbit (approved):
Looks good, thanks!

PR #918, opened 2020-07-13
Refresh clock-related symbolics

lainsce:
Made them follow most newer symbolics, by dropping the 2px thick clock rim in favor of a 1px thick clock rim symbol.
This aligns them with the timer symbolic, for example.

They look like:
Screenshot from 2020-07-13 12 58 08


danrabbit:
Closing as a duplicate of #400

PR #921, opened 2020-07-15
Orange 24px/16px folder-music

lainsce:
This makes it more visible in Dark Mode while making a stronger association with the Music app.


lainsce:
Not needed in favor of #921

waldyrious:
> Not needed in favor of #921

Maybe you meant #925 instead?

lainsce:
Yeah, my bad

PR #909, opened 2020-07-02
Add Inkscape's Calligraphy Tool

Fatih20:
Fixes #714 

This is how the icon look in elementary OS light theme
image
This is how the icon look in elementary OS dark theme
image

There's another layer there that contains things I've used in the process of designing this. I'll be cleaning them up once this is finalized and approved.

PR #924, opened 2020-07-16
Calligraphy cleanup

Fatih20:
Cleaning up of prematurely merged #909.


danrabbit:
Looks like you lost the highlight on the ink. Is that intentional?

Fatih20:
> Looks like you lost the highlight on the ink. Is that intentional?

I fixed in the later commit didn't I?

danrabbit:
@Fatih20 it doesn't look like it. I just branched clean

Fatih20:
Whoops, you're right. There, fixed it.

PR #926, opened 2020-07-16
Make window-close symbolic consistent with window-maximize

lainsce:
This ensures both have the same optical size and styling. It may feel strange at first, but it's because we're accustomed to the tiny non-consistent icon.

Looks like:
close


fabiozaramella:
I'd like to hear what @elementary/ux team think about it. I agree that It looks better once you get used to it.

cassidyjames:
It still feels huge to me. I think this is a case of technical equality versus perceptual equality; because the maximize icon has a thinner part an is only on one axis, it needs to be a bit bigger to feel the same weight as the smaller `x` icon.

So I personally don't love this change, as it makes the `x` button way more prominent. It even feels (almost) as prominent as the toolbar buttons in Files, which I think looks wrong here. But I'll leave this one up to actual icon designers like @danrabbit and @micahilbery.

fabiozaramella:
> It still feels huge to me. I think this is a case of technical equality versus perceptual equality; because the maximize icon has a thinner part an is only on one axis, it needs to be a bit bigger to feel the same weight as the smaller `x` icon.
> 
> So I personally don't love this change, as it makes the `x` button way more prominent. It even feels (almost) as prominent as the toolbar buttons in Files, which I think looks wrong here. But I'll leave this one up to actual icon designers like @danrabbit and @micahilbery.

Mmh i don't agree that it looks as prominent as main buttons but I'd like to hear other opinions too. Maybe we can go the other way round and decrease the size of maximize icon to match.

For reference (it's looks good to me, but obviously i'm not an icon designer, I can only give my opinion):
Schermata del 2020-11-04 20 32 33

lainsce:
I could also go the other way and shrink maximize to the same size as close.

fabiozaramella:
> I could also go the other way and shrink maximize to the same size as close.

Yes, please. Would you mind pushing a PR just to check which option is better?

lainsce:
Made the PR, see above.

waldyrious:
Hey @lainsce — for those of us not in the loop, can you share why this was closed?

lainsce:
Was just something I tried out, and didn't get traction between the core developers/designers, so I just closed it.

PR #927, opened 2020-07-16
folder-videos: Use Videos app metaphor

lainsce:
For consistency OS-wide, I made folder-videos use the metaphor used on the Videos app icon to represent a video.

Looks like this:
folder-videos

In-context:
context


danrabbit:
I'm not sure I like conflating the app icon and folder icons like this

lainsce:
Give it a try for a while. I think this will make it easier for users to know where to find their videos if they open Files.

fitojb:
A folder should look like a folder, no matter the size.

lainsce:
@fitojb -
We use the 24 and 16 sized icons for the Files sidebar currently, not opting for symbolics because of a design decision taken before I was in the team. So the smaller sizes are colored variants of the symbolic as they were before with the filmstrip.

fabiozaramella:
I agree with @lainsce that you get used to it quickly and in my opinion these new icons look more modern than the current ones.

PR #934, opened 2020-08-18
adjust lightening shape

danrabbit:
Screenshot from 2020-08-17 18 11 59@2x

@micahilbery thoughts on making the lightening shape a little closer to the symbolic version?


micahilbery:
I think I'm fine with it. It's a little chonkier but it still looks ok to me. Would we also want to change the power settings icon? That's where I pulled the lightning shape from for the update.

danrabbit:
@micahilbery yeah that sounds good. I'm gonna close this and start a new branch based on new master

PR #925, opened 2020-07-16
Make icons distinguishable under any styling

lainsce:
This will make sure some icons, which had darker elements, work really well in Dark or Light styling of the system.

Sidebar icons look like:
|Light|Dark|
|:---:|:---:|
|light| dark|

PR #942, opened 2020-08-19
Visible Icons V - playlist & library-podcast icons

lainsce:
Breaking up #925, this is the fifth one. Comprises of media-related icons.


fabiozaramella (approved):
Cute and much better with dark style, @elementary/ux any thoughts on this?

PR #944, opened 2020-08-19
Redesign find

### LINKED ISSUE #902: bad contrast for the search icon (magnifying glass) when using dark mode
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
While using dark mode, the contrast of  search icon is bad. Almost every other icon has a dark/slightly dark outline and non-dark color inside for better visibility in every mode. The magnifying glass doesn't have that.

### To Reproduce
1. Open Code
2. Select Dark Mode in Settings
3. Tilt screen

## Platform Information
Screenshot von 2020-06-09 20 24 18

Please check what applies:
- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

micahilbery:
I didn't change much from the original design I mostly scaled it and added edit-find-replace. fixes #902


danrabbit (approved):
Yeah, I don't love love it which is why I didn't propose it, but it does show up way better in dark apps so let's go and we can iterate

PR #958, opened 2020-10-04
Drop 64px and 128px actions

danrabbit:
Something I wanted to do while working on arrows. It doesn't appear that we use these anywhere (with the exception of find-location in Onboarding), so it would be nice to save ourselves a lot of work scaling icons


micahilbery (approved):
If we don't use them anywhere I'm fine with it. :+1:

PR #963, opened 2020-10-09
Add latte palette

micahilbery:


waldyrious:
For reference, here's what it looks like:
Screenshot 2020-10-09 at 10 47 55

And here are the hex codes: `#EFDFC4` `#E7C591` `#CFA25E` `#B6802E` `#804B00`

danirabbit (approved):
Oops didn't mean to never merge this

PR #966, opened 2020-10-23
Remove accessories-screenshot

danrabbit:
This is not an FDO icon name and the screenshot app now uses an RDNN icon name and ships its own icon. 

- [x] The only place this is used is in Switchboard, which is resolved in https://github.com/elementary/switchboard-plug-keyboard/pull/330


micahilbery (approved):
:+1:

PR #973, opened 2020-11-05
Make window-maximize symbolic consistent with window-close

lainsce:
The inverse of #926 for comparison to decide which path to take.

Looks like:
a (1)


fabiozaramella:
Consistent, but tiny... I still think the other option is better.

Issue #979, opened 2020-12-09
Add an icon for Rust (.rs) files


apcamargo:
## Problem
ElementaryOS currently do not have a icon for files with the .rs extension, which are associated with the increasingly popular Rust language.

## Proposal
Add a icon for Rust files. One simple option would be to put the logo over the blank "paper" icon.

[Rust logo license](https://www.rust-lang.org/policies/media-guide):

> The Rust and Cargo logos (bitmap and vector) are owned by Mozilla and distributed under the terms of the Creative Commons Attribution license (CC-BY). This is the most permissive Creative Commons license, and allows reuse and modifications for any purpose. The restrictions are that distributors must “give appropriate credit, provide a link to the license, and indicate if changes were made.” Note that use of these logos, and the Rust and Cargo names, is also governed by trademark; our trademark policy is described below.

## Prior Art

![Screenshot from 2020-12-09 15 43 22](https://user-images.githubusercontent.com/22940964/101672984-4e85c780-3a35-11eb-83ab-0c5d5fc707a8.png)

danirabbit:
Hey thanks for your suggestion. I'm really reluctant to add even more programming language file types. I think we are already supporting too many, especially because a lot of these logos don't look good at small sizes.

I think I would rather have some generic programming language icons that are like "compiled language", "interpreted language", "markup language" and then symlink to these more generic icons. It's unlikely you'll mix programming languages in the same project so it's probably just more useful to figure out like which is a class file and which is like styles or xml or whatever

Gonna close as "Design conflict" since I want to reduce the number of programming language icons rather than expand it

PR #983, opened 2020-12-28
Add Rust mime

apcamargo:
Closes https://github.com/elementary/icons/issues/979


lainsce:
Those small sizes have the Rust logo a lot blurry, could you try to make it fit the pixel-grid?

apcamargo:
@lainsce, I tried to edit the logo to fit into the grid but I couldn't improve it that much. Rust logo is really complex.

Do you think we can proceed this PR with just the 48, 64, and 128 sizes? I've noticed that there are some other icons that are also not available at smaller sizes.

lainsce:
Maybe just 48 and 128 I guess, should cover the divisons pretty well (128 divides to 64 divides to 32 / 48 divides to 16), but try to make those two have the pixel-grid as much as possible.

A followup PR maybe from someone on the UX team may follow :P

apcamargo:
Ok, so I deleted the smaller sizes and improved the fit to the grid a bit more. Rust logo is really complex, so it's not perfect. I did the same for the R logo (https://github.com/elementary/icons/pull/987), although I maintained the 32 version there because I think it is looking good. Thanks again for the tips @lainsce!

In order to get a review from the UX team should I do something, or just wait? I know they have a lot on their plates right now, so I won't tag anyone them here.

lainsce:
Just keep doing PRs and the UX team will go one by one later.

aerkiaga:
Rust's [art license](https://www.rust-lang.org/policies/media-guide) allows modifications to the artwork. I'd suggest making a second, simplified version of the logo that looks good at the smaller sizes. For example, you could remove the four small holes and making the outline a smooth circle with a radius that looks similar to the logo with teeth (radius with no teeth + tooth length/2). Then you could use one version for the larger sizes and another for the smaller ones.

jeremypw:
@danrabbit Has this PR any chance?  It has been outstanding a fair while.

danirabbit:
Hey thanks for your contribution! I just responded to the original issue report. Sorry I know it's not what you're hoping to hear!

>
>

Issue #978, opened 2020-12-05
Add an icon for R files (.R/.r)


apcamargo:
## Problem
ElementaryOS currently do not have a icon for files with the .R or .r extensions, which are associated with the largely used R language.

## Proposal
Add a icon for R files. One simple option would be to put the [logo](https://www.r-project.org/logo/Rlogo.svg) over the blank "paper" icon.

NikoKrause:
Hi @apcamargo,
to my experience you get feedback a lot faster, if you just create a pull request.
Kind regards

PR #987, opened 2021-01-02
Add R mime

apcamargo:
Closes https://github.com/elementary/icons/issues/978

### DIFF
--- a/mimes/128/text-x-r-source.svg
+++ b/mimes/128/text-x-r-source.svg
@@ -1,0 +1,242 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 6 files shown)


lainsce:
Could you go and redo the R logo at the small sizes so it's by the pixel-grid? As it is they are bluury. Thanks. :)

apcamargo:
I'm not very good with that but I can try. Which sizes do you think a redraw is necessary? 16 and 24?

lainsce:
16, 24 and 32 seem to be blurry. Just enable the grid in Inkscape, and try to make straight edges fit the 1px grid the grid provides.

apcamargo:
Thanks for the tips, I'll try! If it still don't look good after that, I'll just make a simplified version with the "R" only.

kurisubrooks:
Based on [this comment](https://github.com/elementary/icons/pull/983#issuecomment-1933321153) by @danirabbit, this PR should be closed.

> Hey thanks for your contribution! I just responded to the original issue report. Sorry I know it's not what you're hoping to hear!
> 
> 
> 

PR #989, opened 2021-01-12
Round symbolic pan-* icons

danrabbit:
Uses the same consistent arrow head shape as the new `go` icons


micahilbery (approved):
:arrow_up_small: :+1:

PR #994, opened 2021-01-15
Add 24px color emblem favorite

danrabbit:
I don't think we're using this anywhere, but `emblem-favorite` is in the fd.o spec and I thought this looked cute


micahilbery (approved):
:heart:

PR #996, opened 2021-01-15
Simply delete the 24px apps icons folder

lainsce:
This PR is self-explanatory. These icons weren't used anywhere in the OS (I personally checked.), and were collecting dust.


danrabbit:
Closing because we do use 24px icons for badges, particularly in notifications

PR #1000, opened 2021-01-23
Fix missing audio-headphones and media-removable

### LINKED ISSUE #999: No icons for audio-headset and media-removable
## What Happened

As title implies.
image

## Expected Behavior

They should have icons.

- `media-removable` can be symlinked to `drive-removable-media-usb`.
- 'audio-headsets` should be renamed to `audio-headset`.

## Platform Information
eOS6

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/96089455-no-icons-for-audio-headset-and-media-removable?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pongloongyeat:
Closes #999

I'm not sure renaming "headsets" to "headset" here would be a good idea since some applications may already be using the icon name "headsets" so I symlinked it instead. Let me know if this needs to be reverted.

`audio-headset` and `media-removable`:
image
image

`drive-removable-media-usb`:
image
image


danrabbit (changes requested):
It looks like you've used symlinks that point to locations on your specific filesystem, so they won't work on another computer.

danrabbit:
It doesn't look like `audio-headsets` is being used anywhere so it should be safe to use `git mv` to rename to `audio-headset`

pongloongyeat:
> It looks like you've used symlinks that point to locations on your specific filesystem, so they won't work on another computer.

Ahhhhh you're right I didn't even notice that. Thanks for pointing it out!

> It doesn't look like `audio-headsets` is being used anywhere so it should be safe to use `git mv` to rename to `audio-headset`

I'm more worried about external curated apps which may use the current icon name. Should this be a concern?

danrabbit:
@pongloongyeat No, icons won't be released to 5.x anymore and developers will expect API breaks between 5.x and 6.x

danrabbit (approved):

PR #1004, opened 2021-02-23
bigger 24px battery symbolic

danrabbit:
I'm not sure if this is ridiculous or if it actually makes more sense. Drafting for feedback

Screenshot from 2021-02-23 10 35 09@2x


cassidyjames:
Well now I can't unsee the battery looking kind of small in the existing icon set. :thinking: I think this would be worth giving a try.

danrabbit:
the more I look at it, the dumber this feels lol Maybe a middle ground

danrabbit:
Screenshot from 2021-02-23 11 07 55@2x

Maybe a little better

danrabbit:
Screenshot from 2021-02-23 16 46 20@2x

Starting to get into the bigger size

cassidyjames:
@danrabbit yeah I like this middle ground!

micahilbery (approved):
:+1: :battery:

PR #1008, opened 2021-02-26
Update the look of SSD and USB drives

hanaral:
Updates the look of both SSD and removable USB drive icons

ssd128

SSD:
- [x] 16px
- [x] 24px
- [x] 32px
- [x] 48px
- [x] 64px
- [x] 128px

USB:

- [ ] 16px
- [ ] 24px
- [ ] 32px
- [ ] 48px
- [x] 64px
- [ ] 128px

I've followed the HIG and tried to keep it within elementary's style - Though I still feel I should add something to the front of the USB drive, since it looks quite barren


hanaral:
@micahilbery @danrabbit sorry to tag you both like this, but I wanted to discuss a couple of things;
- I'm wondering if I should try flattening the perspective at smallest sizes for visibility, but looking at macOS on the iMac 5k makes me think that it won't be a problem
- I also want to use a custom version of the USB flash disk icon for USB 3 devices, I assume Files _can_ tell the difference between 2.0 and 3.0 
(I could cheat and give it to the removable HDD assuming it'll be 3.0 - though that might create issues)

PR #1022, opened 2021-04-15
Emblems: Add symbolic version of Files' emblems

danrabbit:
Screenshot from 2021-04-15 13 17 25


micahilbery (approved):
Looks Good!

hanaral:
Would there be no chance of getting the old padlock symbolic back and re-doing the no-entry and link icons in the same style? I Think having a two-tone look makes them look nicer and less like badges (they can't be badges because they aren't on top of another icon, right?)
Sorry to post on a finished PR!

PR #1025, opened 2021-04-27
Add git emblems

danrabbit:
In the interest of https://github.com/elementary/files/issues/1675


jeremypw:
Icon View
Screenshot from 2021-04-27 10 57 58

List View
Screenshot from 2021-04-27 10 58 43

jeremypw:
The `emblem-git-modified` emblem appears paler in IconView (not symbolic) than in ListView (symbolic). In IconView the contrast with the color of the folder is not so great.

cassidyjames:
I wonder if we can use more than color to show the difference between new and modified? Like some sort of `+` for new? I know there's not much space to work with here, though.

danrabbit:
Yeah it's kind of hard to convey both the idea of git and also the status in such a small space

cassidyjames (approved):

PR #1042, opened 2021-09-14
Remove OARS icons

danrabbit:
Compliment to https://github.com/elementary/appcenter/pull/1667

Since these are only used in AppCenter, it makes more sense to maintain them there


cassidyjames:
I think the intent was to use them in the Screen Time & Limits plug, but in the interest of YAGNI I'm fine moving them, at least for now.

danrabbit:
Yeah I think it would just be easier to iterate on in AppCenter for now and if we end up implementing that in System Settings then we can move the icons around or copy them or whatever makes more sense

PR #1043, opened 2021-09-23
Rounded audio-volume

### LINKED ISSUE #1041: Make wingpanel-indicator-sound rounded
## Problem

It's one of the only icons left that still looks really square. The speaker itself could also be longer on the left, kind of like Apple's volume icon in MacOS Big Sur and iOS 14.

## Proposal

Make a new, rounder volume button to go with the rounder look of the OS.

## Prior Art

iOS 14:
screen-2

MacOS Big Sur:
Screenshot from 2021-08-31 19 58 18

danrabbit:
Fixes #1041

Been staring at these too long to know if this is better or worse :sweat_smile:


micahilbery:
Fwiw I like these better but they should also probably accompany a redesigned wifi indicator icon.

micahilbery (approved):
Love it! ❤️

PR #1048, opened 2021-10-20
Remove Evince icons

danrabbit:
We're now shipping this icons in our Evince Flatpak

This isn't an fd.o icon name and we're not using it anywhere else


meisenzahl (approved):
Sounds sane 😃

PR #1049, opened 2021-10-20
Remove FileRoller icons

danrabbit:
Depends on https://github.com/elementary/fileroller/pull/7 and shipping the FileRoller flatpak

This isn't an fd.o icon name and we don't use it anywhere else


meisenzahl:
@danrabbit should be ready for review as https://github.com/elementary/fileroller/pull/7 was merged.

PR #1052, opened 2021-11-05
Redesign PDF file type

danrabbit:
Based on https://github.com/elementary/evince/pull/28


micahilbery (approved):
Keeps it consistent with the application it's related to. So it's a :+1: from me.

cassidyjames (approved):

PR #1053, opened 2021-11-05
Add Font preferences

danrabbit:
Differentiate between the font viewer app icon and font preferences

applications-fonts is not an fd.o icon name and afaict it is unused


cassidyjames (approved):
Makes sense!

PR #1056, opened 2021-11-09
Actions: Remove arrows from Copy and Paste

danrabbit:
Symbolic versions don't have an arrow and they seem still to convey their meaning. Also helps reduce noise in uses like: https://github.com/elementary/terminal/pull/639

And saves me having to update an arrow shape yay lol


micahilbery (approved):
I agree that this still gets the meaning across and makes it less busy. :+1:

PR #1061, opened 2021-11-23
Remove apps/multimedia-photo-manager

danrabbit:
Depends on:
- [x] https://github.com/elementary/photos/pull/670

This isn't an fd.o icon name and we ship this icon with Photos as RDNN now


micahilbery (approved):
:+1:

PR #1072, opened 2022-01-19
Redesign media-playlist-repeat-symbolic

danrabbit:
Goals:
* Use the slashy for more clear disabled state
* Increase legibility for each state by making sure elements don't overlap arrows
* Use new rounded arrow shape

Screenshot from 2022-01-19 11 39 24

Some comparison between old shape and a couple new proposals

Screenshot from 2022-01-19 11 31 15


cassidyjames (approved):
🔥

PR #1074, opened 2022-01-21
Status: update shuffle and consecutive

danrabbit:
Use new arrow shape and match size of repeat:

Screenshot from 2022-01-21 13 23 46

and in context:

Screenshot from 2022-01-21 13 27 19


micahilbery (approved):
Looks great! :+1:

Issue #548, opened 2018-05-13
mimes missing sizes:


NikoKrause:
a few icon sizes are missing for:

- [ ] image-x-xcf
- [ ] authors
- [ ] image-vnd.adobe.photoshop
- [ ] font-x-generic
- [ ] extension
- [ ] audio-x-playlist-rtl
- [ ] text-x-generic-template-rtl

PR #1079, opened 2022-01-27
Add/update eBook and eDocument mimetype icons

### LINKED ISSUE #1075: Missing comic book mimetype icons
### Problem

Mimetype icons are missing for `.cbz` and `.cbr` files, mimetypes `application/vnd.comicbook+cbz` and `application/vnd.comicbook-cbr`.

These are zip and rar archive files with images inside. They can be read by almost any ebook or pdf reader (Evince, Okular, Calibre, etc). Right now they just default to the document icon.

### Proposal

Possible solutions:

1) Symlink to archive icon
2) Symlink to ebook icon
3) Make a new unique icon

They are archive files but, like epub/ebook, they are books within an archive file so 2 might make a little more sense since they are technically electronic books. If 3, I don't know if I could make an icon that meets the standards for this theme but I could try if 1 and 2 aren't good solutions. Maybe the ebook icon with a speech bubble or something would be simple enough.

### Prior Art (Optional)

_No response_

newhoa:
I'd like to add missing comic book and possibly other ebook mimetype icons. I'd really appreciate input/direction.

Currently this is what I have as the comic book icon:
(left current epub/ebook, right proposed comic book icon)

comic-icon-rc

Just a slight modification of the eBook icon. Does `?!` work or should those symbols be reserved for dialog icons? Is it too silly looking compared to the more serious/professional look of the rest of the icons?

A few other eBook icons are also missing. Here is a list and proposals.

* Amazon eBook and Mobipocket icons (`.azw`, `.azw3`, `.mobi`) -> symlink to epub/ebook icon
* Deja Vu (`.djvu`) - > symlink to pdf icon, this is a scanned document format like pdf or ps
* PostScript (`.ps`) -> symlink to pdf (edocument instead of ebook but just noticed it is missing)

Also, the FB2 icon is missing at 128px. I can add this but before I do -- should I keep the orange look and the `fb` lettering? Or switch to the green (and possibly the e of the epub/ebook) so all ebook file icons consistent?

comic-icon-fb2

Fixes #1075 
Addresses part of issue #548 

Thanks!

### DIFF
--- a/mimes/128/application-postscript.svg
+++ b/mimes/128/application-postscript.svg
@@ -1,0 +1,1 @@
+application-pdf.svg
--- a/mimes/128/application-vnd.comicbook+zip.svg
+++ b/mimes/128/application-vnd.comicbook+zip.svg
@@ -1,0 +1,388 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   version="1.1"
+   width="128"
+   height="128"
+   id="svg4134"
+   sodipodi:docname="application-vnd.comicbook+zip.svg"
+   inkscape:version="1.0.2 (e86c870879, 2021-01-15)">
+  <path
+     style="display:inline;fill:url(#linearGradient4333-3);fill-opacity:1;stroke:url(#linearGradient1006);stroke-width:1;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:0;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1"
+     id="path2723"
+     d="m 108.5,7.5 v -3 c 0,-1.941931 -0.84186,-3 -4,-3 h -81 v 6"
+     sodipodi:nodetypes="ccccc" />
+  <path
+     style="color:#000000;display:inline;overflow:visible;visibility:visible;fill:url(#linearGradient3156);fill-opacity:1;fill-rule:nonzero;stroke:url(#linearGradient3158);stroke-width:1;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;marker:none;enable-background:accumulate"
+     id="rect5505-21-3-9"
+     d="m 27.5,7.5 h -10 V 4.08617 C 17.5,2.150302 18.144001,1.5 20.046842,1.5 H 27.5"
+     sodipodi:nodetypes="ccssc" />
+  <sodipodi:namedview
+     id="namedview71"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
+     inkscape:pagecheckerboard="0"
+     showgrid="false"
+     inkscape:zoom="26.193818"
+     inkscape:cx="59.565546"
+     inkscape:cy="79.941552"
+     inkscape:window-width="1319"
+     inkscape:window-height="954"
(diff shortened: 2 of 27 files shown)


waldyrious:
> Amazon eBook and Mobipocket icons (`.azw`, `.azw3`, `.mobi`) -> symlink to epub/ebook icon

Since `.epub` is a specific format (and the icon in the epub/ebook icon is the [EPUB symbol](https://en.wikipedia.org/wiki/EPUB)), IMHO there should be a generic ebook icon that doesn't use any specific symbol/logo.

newhoa:
> > Amazon eBook and Mobipocket icons (`.azw`, `.azw3`, `.mobi`) -> symlink to epub/ebook icon
> 
> Since `.epub` is a specific format (and the icon in the epub/ebook icon is the [EPUB symbol](https://en.wikipedia.org/wiki/EPUB)), IMHO there should be a generic ebook icon that doesn't use any specific symbol/logo.

Makes sense, thanks! Ok, I updated. Currently:

- Comic icon (`.cbz`, `.cbr`) eBook -> eBook with `?!` speech bubble
- Amazon and Mobipocket eBook (`.azw`, `.azw3`, `.mobi`) -> generic eBook icon
- Deja Vu (`.djvu`) eDocument - > symlink to pdf icon
- PostScript (`.ps`) eDocument -> symlink to pdf icon
- Fictionbook (`.fb2`) eBook -> added 128px icon, propose making it green to match other eBooks

For Mobipocket/Amazon, currently using blank green book/ebook icon (right), could try adding text or a bookmark (left), but bookmark may make it feel like a bookmark file. So might be best to leave it blank/generic unless someone can suggest something that symbolizes Amazon/Mobi (can't use their logo of course).
e-icon-book-generic

FictionBook/fb icon, added 128px. I like the green better, matches the other eBooks if the above changes go through. If green is preferred I can change the other sizes as well.
e-icon-fb

Any more input would be helpful. Thanks!

danrabbit:
Hey thanks for looking into this!

I would lean away from changing the file type icons for any formats that aren't supported out of the box. It's pretty easy for file types to get out of control trying to support absolutely every format, so we've really been slimming these down. Anything in the `application-x-*` namespace should technically be handled by those applications when they're installed, so we should avoid touching those if possible.

In fact, it would be worth probably dropping some of the filetype icons in the current master. I would be inclined to remove the fictionbook icons tbh. But if it stays, the fictionbook icon should remain orange since that's the brand color used in the fictionbook logo.

Something we should test is if fallbacks in the namespace work for these `+zip` etc names. For example, instead of symlinking `application-vnd.comicbook+zip` and `application-vnd.comicbook-rar`, you might be able to do `application-vnd.comicbook` and we should fall back correctly there to the less specific namespace

For the comicbook icon I like the metaphor, but I might try to re-use the bubble shape from `internet-chat` in apps. And for the text it would probably help for hinting not to use italic and use Inter Bold as the typeface

newhoa:
@danrabbit, What would you suggest about the other eDocument/eBook mimetypes in this PR. DejaVu and PostScript are both read by Evince. Is using the pdf/eDocument icon okay for those? Mobi/AZW are not read by default (but neither is ePub). Use the generic eBook icon or don't worry about those?

> Anything in the application-x-* namespace should technically be handled by those applications when they're installed...

Is this an elementary convention or a standard for mimetypes? I don't know much about the naming scheme of mimetypes and the reasoning behind some of them. I've noticed some mimetype icons are provided by an app if it can only be opened with that app, `application-x-ardour` for example. But then you have a format like `tar` which uses `application-x-tar` which no app seems to provide an icon for and can be opened by many apps. That's an old format but there are a lot of modern ones to which the same applies, or for which the official app is not available on Linux though the format can be read by plenty of other apps. And unfortunately when files don't have icons or have incorrect icons, to the end user who doesn't know the reasoning, it feels a little incomplete or the fault of the system/theme.

For the comic icon, I tried using the name `application-vnd.comicbook` and it did not seem to work. That would've been really nice.

I tried this version with the `internet-chat` icon (with and without outline):
e-icon-book-chat-inter

I also tried an outline on the first version, but the `ePub` `e` (and other lettered icons) does not have a border so it isn't very consistent.
e-icon-book-comic-outline

Personally I feel the chat bubble should be reserved for chat related icons (as it is also used for the Iridium app icon which could be confusing). Usually comic bubbles have a longer sharper tail. And the font is usually somewhat italic, with question and exclamation marks usually being bold and more italicized. [Here's an example](https://user-images.githubusercontent.com/1984060/152747236-03794675-17c8-420d-b1a4-485018feb2b3.jpg). That was kind of my rationale for the original icon. With the chat bubble and normal font it looks a little like a chat or help symbol to me. But I've been playing with it too long so my perspective might be way off.

Anyway, just let me know how (or if) I should continue with this. I get adding a ton of mimetype icons can be a headache for upkeep and everything and may go against the simplicity ethos so I don't want to throw stuff in that will complicate things. Just let me know. Thanks!

newhoa:
Closing unless I get any feedback/suggestion to move forward.

PR #1080, opened 2022-02-01
Update deb package icon

newhoa:
Edit: This PR should be ready. As discussed below, this change is now only for the `deb` icon.

---

With the new flatpak and archive icons, I thought it would be nice to update the `deb` icon and make it look more like a package/archive than a document. Then figured while I was at it, might as well add icons for `rpm` and `appimage` packages. They all use the new flatpack package icon as a base.

Proposed:
e-icon-packages

I'm also considering adding a `snap` package icon, trying to figure out if the snap logo is GPL compatible.

If these look okay, I'll add the other sizes. Thanks.


danrabbit:
Hey thanks for looking at this. I'd rather not include icons for packages that aren't supported by elementary OS here. Flatpak is really the only package users can install in a graphical way. I think I'd be okay with changing the deb icon since those can be installed via terminal, but RPM and AppImage wouldn't be accepted here I don't think

waldyrious:
I can understand not adding RPG and Snap icons, but AppImages, being self-contained and not depending on any system-wide changes, would seem neutral enough IMHO.

danrabbit:
I _really_ don't want to make it look like we support AppImages. I think it's important that they look unsupported

waldyrious:
Ok, I disagree with making something look like it's unsupported when it technically works without any previous preparation and without affecting the rest of the system, but I can see where you're coming from. Thanks for clarifying.

newhoa:
I'll stick with just the `deb` unless something else is decided.

For 16px, should I go with the package or leave it just the deb logo? The package is more consistent but less legible.

e-icon-packages-16


Also, I notice `.snap` and `.apk` currently use the empty document. Should I at least symlink those to the generic archive icon since they're compressed packages/archives?

danrabbit:
@newhoa Let's go with the package for consistency here

I would recommend addressing that in a separate pull request so we don't cross concerns here, but a hesitation I have is that since those aren't handled by elementary OS, providing an icon for them would I think overwrite the icon provided by an app that would handle them. Generally, it's probably better to ignore file types that aren't supported out of the box

danrabbit (approved):
Apologies for taking a while to review this. Looks good, thank you!

PR #1081, opened 2022-02-01
Add extension mimetype at 128px

newhoa:
Adds extension mimetype and related symlinks at 128px.

Addresses part of issue #548


danrabbit (approved):
Looks good, nice work!

PR #1082, opened 2022-02-02
Add 128px icons for initial-setup

danrabbit:
Drafting because probably wanna do installer too


waldyrious:
Just out of curiosity, why is this key wider than the ones to the left?

Screenshot from 2022-02-02 11-10-39

danrabbit:
Pixel fitting :)

waldyrious:
> Pixel fitting :)

Thanks for the information! Would the pixel alignment still be possible by widening the backspace key instead?

micahilbery (approved):

danrabbit:
@waldyrious I'm going to merge this for now so it's not blocking other stuff, but can you open an issue report about that backspace key so we don't forget?

PR #1085, opened 2022-02-19
Update media stop icon

newhoa:
Changes the `media-playback-stop` icon to square. Should now match the symbolic icon and better match the newer style of the other media control icons.

I also noticed the `media-playlist-repeat` icon was missing at 24px so I added that. If that should be removed let me know.

Thanks.

media-stop-compare1

Edit: Adjusted lighting of the repeat icon, updated pic.


newhoa:
Any feedback on this one? I feel like the current stop icon resembles a record button and doesn't match the symbolic version. Is the red still too danger/record-like? I tried some other colors:

stop-colors

Don't know if any of those are any better. Any feedback would be appreciated.

danrabbit (approved):
Looks good, thank you!

PR #1087, opened 2022-02-21
Use rounder edges in template and download folders

newhoa:
Wasn't sure if anyone had started adding the new rounder arrow style to the `Download` folder so I started on it.

Then I got distracted by the `Templates` folder icon and wanted it to look more similar to the others with smoother edges.

Let me know if either of these look okay and I can start working on the other sizes.

Thanks!

Current on left, proposed on right
folder-templates-final1

folder-download-final1

Edit: I realized `folder-remote` could use the rounder arrow update. Let me know if you all want me to do this one too or if it's already been/being done somewhere.

folder-recent-1a

---

Downloads folder arrow should now match the new arrow style.

Ruler on the templates folder given rounded edges
to match rounded edges of other folder icons.
Added hatch marks for a more distinct silhouette
and more similar to ruler used in the mimetypes.

### DIFF
--- a/places/128/folder-download-open.svg
+++ b/places/128/folder-download-open.svg
@@ -1,17 +1,39 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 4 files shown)


danrabbit:
Nice updates! I'd generally recommend splitting things into smaller pull requests so that one thing doesn't get held up by another, but this is a good direction :)

newhoa:
Thanks for the feedback! I'm closing this and submitting these proposals individually.

PR #1090, opened 2022-03-22
Use rounded arrow on download folder

newhoa:
Updated the places folder-download folders to use the rounded arrow style. Tried to also make the smaller sizes a little more legible and consistent while I was at it.

---
Current:

current-white

Proposed:

proposed-final2-white


micahilbery (approved):
:heart::+1:

PR #1091, opened 2022-03-22
Fix incorrectly named pgp/gpg mimetype

newhoa:


danrabbit:
Hm, have you confirmed that this is broken? These icon names should work because of fallback naming

newhoa:
> Hm, have you confirmed that this is broken? These icon names should work because of fallback naming

It doesn't work for me, maybe you or someone can double check (can generate by doing `gpg --gen-key` and look in the `.gnupg` folder). Maybe I've fiddled around with my setup too much. But you're right, it should work. If it doesn't I guess it's a Files or Gtk/Gdk bug. If anyone can confirm this is a problem let me know and I can file an issue elsewhere. Till then I'll close this. Thanks.

danrabbit:
I see what's happening here. It looks like there's two icon names we want to cover here so this was named assuming the icon naming fallbacks were working. I proposed #1125 to add symlinks since that's not currently functioning

PR #1096, opened 2022-04-20
Devices: rounder network squares

danrabbit:
In line with the border radius used in the new Music icon and in previous branches

Update green icon color palette while we're here


micahilbery (approved):
:+1:

PR #1097, opened 2022-04-20
Invert updates arrow

danrabbit:
This has annoyed the hell out of me. They're updates, not "down"dates


micahilbery (approved):
:arrow_up:

PR #1106, opened 2022-05-19
Clean up prefs-desktop-workspaces

danrabbit:
Use the new square format, make sure we have consistency throughout


micahilbery (approved):
:fire:

PR #1110, opened 2022-06-06
Categories: update preferences-desktop-keyboard

danrabbit:
Updated to use the new square shape and Inter as the typeface

~~Drafting because I'm hoping for some feedback from -rtl users: https://twitter.com/DaniElainaFore/status/1533618566615793666~~

It sounds like according to RTL users that the @ symbol is universal and doesn't get flipped in RTL, so I just adjusted the position

I changed the character position because it works better if the icon is badged:
Screenshot from 2022-06-05 18 39 41


danrabbit:
Alternatively, the looped square is a key we use for shortcuts :shrug:. Any thoughts @micahilbery?

preferences-desktop-keyboard

micahilbery (approved):

PR #1111, opened 2022-06-16
FileTypes: Use rounded square for audio icon

newhoa:
The audio mimetype icon currently has the look of an icon that would represent a document or text. Most mimetype icons that represent binary files use a different shape (Package, Video, Executable files use a more square shape).

This switches the shape of the audio icon to more closely resemble the style of other binary mimetypes and makes the audio icon look less like a document.

I used the new `image-missing` icon and adjusted the gradient, borders, and shadows to match the current audio icon. I think this gives it a mimetype look while the shape differentiates it from a file that represents text.

### DIFF
--- a/mimes/128/audio-x-generic.svg
+++ b/mimes/128/audio-x-generic.svg
@@ -1,134 +1,94 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:xlink="http://www.w3.org/1999/xlink"
-   version="1.1"
-   width="128"
+   id="svg4434"
    height="128"
-   id="svg3172">
+   width="128"
+   version="1.1"
+   sodipodi:docname="audio-x-generic.svg"
+   inkscape:version="1.1.2 (0a00cf5339, 2022-02-04)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:dc="http://purl.org/dc/elements/1.1/">
+  <sodipodi:namedview
+     id="namedview42"
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1.0"
+     inkscape:pageshadow="2"
+     inkscape:pageopacity="0.0"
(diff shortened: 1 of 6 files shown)


danrabbit:
I'm a little torn on this one. I'm almost more inclined to go the other direction and make sure some of the other mimetypes like video, photos, etc are using the document shape. I'd kind of rather use the squares to represent things that are either applications or parts of applications as opposed to files

newhoa:
I was a little concerned about it looking too much like an app icon. I was hoping that by using the same gradient style as the other mime icons that it would look enough like a mimetype to not be confused with an app icon. Especially since they are to be seen in context with other mime icons.

This is similar to what was done with flatpak and flatpakref icons. It is easy to differentiate between what is text and what is not.

I do think that using the document/sheet of paper style and aspect ratio for things that aren't a document or that don't represent a text-based file is a little confusing.

The document/paper shape and ratio is a visual cue that I think to most would indicate text, especially when some of them (likely the most common ones) do have text on them. So when you see a document shape with text on it and then a document shape with a music note, it is difficult to understand that when the first one represents something that will open as text-readable that the second one will not do the same. This gets more confusing when there are icons like markdown, vala, etc that do not have text on them but are still text-readable files -- the document/paper shape still indicates that they are text files which is correct. So the fact that they have visuals on them that aren't text is still okay because the icon shape itself still indicates a document. But for audio, video, pictures, packages, etc this doesn't make a lot of sense.

Another example would be the subtitle and video file. If they both look like documents, it's difficult to tell that one is text and the other is not. Same with the playlist and audio icons. I think the way you can distinguish between flatpak and flatpakref is a good way to address it.

So I think it would be helpful to have a different shape and/or look for binary, non-text-readable icons.

newhoa:
Here were some other styles I played with (I'll refer to these as icons 1 thru 6):

audio-prop2

1 - Current document icon
2 - Current document icon horizontal
3 - Current document icon horizontal with 4px corners (this is what video uses)
4 - Executable/Package icon shape/size with square corners like current document icon
5 - Executable/Package icon shape/size with 4px corners (video also uses 4px)
6 - Executable/Package icon shape/size with 10px rounded corners

Not sure if any of these would be better or still not the right direction.

danrabbit:
@micahilbery thoughts on the above? ^

newhoa:
The audio one is very similar to the [older audio icons](https://github.com/elementary/icons/commit/a83ddad7c19433e96bb8ae839de56e20db8398cf), which I always really liked! If it's okay to reintroduce that style I think that's a nice way to do it.

For the mockups, I'm torn between row one and three for text and non-text files (although I prefer the executable/package rounding in row 2, I don't like the heavily rounded borders for document and non-text files.). Rounded borders of documents in row 3 looks a bit odd to me. But at the same time, sharp borders on non-text files in row 1 also looks a bit odd to me. But like you said, could just be that it's something that takes getting used to.

How would you feel about a progressive rounding based on the type of file?

audio-prop3

* Document, sharp like paper.
* Binary files, slightly rounded to differentiate them from documents but not rounded enough to look like app icons.
* Executables and packages (which can also be executable/installable like flatpak, deb, etc) be more rounded since they could themselves be or represent apps.

I think this gives good visual clarity. And it would be the most practical to implement as it would minimize the amount of files that needed to be reworked.

newhoa:
I did a few in this style and in different colors to see how it would look once more were added. I do like the idea, my only concern is that they may be a bit loud/distracting and may overshadow other icons. But it could just take some time to adjust. Gonna spend some time living with them. Just thought I'd attach them here in case anyone else wants to try them out.

new-non-text-mimes-light

[new-non-text-file-mimes.tar.gz](https://github.com/elementary/icons/files/8933317/new-non-text-file-mimes.tar.gz)

danrabbit:
I guess another thing to consider is that there are some non-text file icons that use more than one color in their symbolics

font-x-generic

Issue #578, opened 2018-05-17
Inconsistency of symbolic links; build them on runtime


NikoKrause:
I'd like to make a suggestion.
I noticed that there are symbolic links in some sizes, which are not in there in other sizes, e.g. there is a symbolic link `apps/16/calendar.svg` but not `apps/24/calendar.svg`.

My suggestion would be to make a list for symbolic links, something like
```
image-x-generic.png <- image.png
image-x-xcf.png <- image-x-compressed-xcf.png
internet-feed.png <- application-atom+xml.png
internet-feed.png <- application-rss+xml.png
model.png <- application-x-blender.png
model.png <- image-x-3ds.png
office-contact.png <- text-directory.png
office-contact.png <- vcard.png
package-x-generic.png <- package.png
```
and create the symbolic links on build time. This includes to remove the symbolic links from the repo and just keep the actual svg files, which will make everything look clearer (also no more vacuum svg files, if you only add symbolic links -> cleaner git history).

PR #1115, opened 2022-06-17
Mime links

danrabbit:
In the interest of #578

Creates all the symlinks for file types using the build system. Some mimes are links to icons outside this directory, so I didn't touch those. I also left the Flash mime alone because I'm going to propose its removal in another branch


micahilbery (approved):
Makes sense to me :+1:

PR #1118, opened 2022-06-23
Mimes: redesign video-x-generic

danrabbit:
In the interest of some of the discussion on #1111 

* Use the same background color as other mime icons
* Replace the filmstrip symbolic with the "play" icon which is less skueomorphic and more closely related to the videos app icon


tintou:
I don't know how to address this, but to me the new version almost looks clickable and not enough like a medium :shrug: 

Here were my unsolicited 2c

PR #1128, opened 2022-08-24
Apps: redesign application-default-icon

danrabbit:
The application-default-icon redesign from #1076 with the more rounded square and using a different metaphor and color

### DIFF
--- a/apps/128/application-default-icon.svg
+++ b/apps/128/application-default-icon.svg
@@ -13,24 +13,15 @@
   <defs
      id="defs4381">
     <linearGradient
-       gradientTransform="matrix(0.89553576,0,0,-0.89663265,6.6836208,69.705837)"
-       gradientUnits="userSpaceOnUse"
-       xlink:href="#linearGradient4113"
-       id="linearGradient3149"
-       y2="3.4508991"
-       x2="76.523361"
-       y1="-9.023138"
-       x1="76.523361" />
-    <linearGradient
-       id="linearGradient4113">
+       id="linearGradient880">
       <stop
+         style="stop-color:#cd9ef7;stop-opacity:1"
          offset="0"
-         style="stop-color:#000000;stop-opacity:1"
-         id="stop4115" />
+         id="stop876" />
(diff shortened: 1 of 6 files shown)


micahilbery (approved):
:star:

waldyrious:
I'm having a hard time figuring out what the intended metaphor for the star is. As a user, I would be confused by what the symbol is supposed to mean, since it's normally associated with favorites or ratings. Could you share what was the reasoning behind the choice?

danrabbit:
@waldyrious just something I'm trying out. I don't feel like we have a great metaphor to represent applications. This icon is mainly used to represent a generic application or as a fallback for when an app icon fails to load or is missing. I'm definitely open to other metaphor options! Could you open a new discussion or issue report for that though? Closed pull requests aren't a great place for new discussions

waldyrious:
I don't really have an alternative proposal to make, other than perhaps the former symbol which IMHO seemed somehow more appropriate (i.e. the gear representing some sort of engine for _doing_ something). Would an issue suggesting a return to the old metaphor, or simply asking for alternative ideas without proposing one upfront, be acceptable?

jeremypw:
Maybe we could provide more generic icons for other AppStream.ComponentKinds?  At the moment we only accomodate `ComponentKind.ADDON` in AppCenter.  If we can reduce the number of occasions we are forced to use the last resort fallback then the less important it becomes.

danrabbit:
@waldyrious yep that would be fine, thanks!

waldyrious:
Sorry I never got around to open the issue :sweat_smile: T be honest, I was still hesitant to request a change without providing a concrete direction other than a mere reversal.

In any case, I suppose the change in #1135 (changing [`application-x-executable`](https://github.com/elementary/icons/blob/master/mimes/128/application-x-executable.svg) to the gear icon, rather than a symlink to `application-default-icon` which was converted to the star metaphor here) would somewhat reduce the impact of this change, so I'd be fine with the situation assuming that PR is merged.

PR #1129, opened 2022-09-25
Update symbolic bookmark icons

danirabbit:
* Consistently use nice rounded star
* Remove `bookmark-missing` and `star-new` which are unused and not in fd.o


micahilbery (approved):
:star:

PR #1132, opened 2022-09-27
Round actions/document-* symbolic

danirabbit:
Round out these shapes a bit more and fix some inconsistencies with color versions


micahilbery (approved):
love it!

PR #1134, opened 2022-09-28
Categories: symlink preferences-desktop-wallpaper

danirabbit:
I want to give this its own icon in a future version, but for now symlink so that we don't break anything that's relying on this icon name


micahilbery (approved):
👍

PR #1135, opened 2022-09-28
Modernize 128px application-x-executable

danirabbit:
Make sure we use a different icon for binary file types vs fallback app icons. Use palette colors. Scale down "Windows" symbol


micahilbery (approved):
Looks good! :+1:

PR #1137, opened 2022-09-29
Status: Update headphones symbolic

danirabbit:
* Create a 24px icon that's scaled closer to other panel icons
* Use more pronounced slash like the new sound icons


micahilbery (approved):
subtle but very welcome change to the design. :+1:

PR #1139, opened 2022-10-01
Fix meson warning for run_command

turbobastii:
When running meson build, there is a warning about run_command changing its defaults in the future.
(See: https://github.com/mesonbuild/meson/issues/9300)

The same issue was addressed in this PR: https://github.com/elementary/code/pull/1207

This fixes the warning while keeping the current default.


cursors/meson.build
```diff
@@ -94,7 +94,8 @@ foreach size: sizes
                 rsvg, '--format', 'png',
                 '--width=@0@'.format(size), '--height=@0@'.format(size),
                 '--output', '@0@.png'.format(join_paths (meson.build_root(), 'cursors', size + 'x' + size, cursor)),
-                join_paths ('24', cursor + '.svg')
+                join_paths ('24', cursor + '.svg'),
+                check: false
```
eli-schwartz, line 98: It's unfortunate that build outputs don't use custom_target, but using `check: false` is absolutely wrong for this, because if rsvg fails here then that's very bad news and the build is broken.

cursors/meson.build
```diff
@@ -105,7 +106,8 @@ foreach size: sizes
                 rsvg, '--format', 'png',
                 '--width=@0@'.format(size), '--height=@0@'.format(size),
                 '--output', '@0@.png'.format(join_paths (meson.build_root(), 'cursors', size + 'x' + size, watch)),
-                join_paths ('24', watch + '.svg')
+                join_paths ('24', watch + '.svg'),
+                check: false
```
eli-schwartz, line 110: ... and here

cursors/meson.build
```diff
@@ -122,7 +124,8 @@ foreach cursor : cursors
         xcursorgen,
         cursor + '.in',
         '-p', join_paths (meson.build_root(), 'cursors'),
-        join_paths (meson.build_root(), 'cursors', cursor)
+        join_paths (meson.build_root(), 'cursors', cursor),
+        check: false
```
eli-schwartz, line 128: ... and here.

eli-schwartz (commented):
> This fixes the warning while keeping the current default.

Please don't keep the current default, the entire reason for the warning is because the current default is usually bad, and in this particular case, it goes to show why that was bad.

meson.build
```diff
@@ -91,7 +91,7 @@ foreach directory_entry : directories
     directory = directory_entry[0]
     directory_context = directory_entry[1]
     foreach size: sizes
-        result = run_command(ls, join_paths(meson.source_root (), directory, size))
+        result = run_command(ls, join_paths(meson.source_root (), directory, size), check: false)
```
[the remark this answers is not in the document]
turbobastii: Thanks for pointing it out. I replaced it with the is_dir call and verified, that it works as expected (as good as I could, as a meson beginner).

cursors/meson.build
```diff
@@ -85,7 +85,7 @@ sizes = [
 ]
 
 foreach size: sizes
-    run_command('mkdir', join_paths (meson.build_root(), 'cursors', size + 'x' + size))
+    run_command('mkdir', join_paths (meson.build_root(), 'cursors', size + 'x' + size), check: false)
```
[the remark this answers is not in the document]
turbobastii: Thanks, changed the check to true.

cursors/meson.build
turbobastii: Thanks, changed the check to true.

cursors/meson.build
turbobastii: ...changed it :D

cursors/meson.build
turbobastii: ...changed it :D

meson.build
```diff
@@ -91,8 +92,8 @@ foreach directory_entry : directories
     directory = directory_entry[0]
     directory_context = directory_entry[1]
     foreach size: sizes
-        result = run_command(ls, join_paths(meson.source_root (), directory, size), check: true)
-        if (result.returncode() == 0)
+        dir_exists = fs.is_dir(join_paths(meson.source_root (), directory, size))
+        if (dir_exists == true)
```
eli-schwartz, line 96:
```
dir_exists = true
if dir_exists == true
```

simplifies to `if <bool> == true` and also (because the return value of `<bool> == true` is another `<bool>` with the same value as the left hand side) to just `if <bool>`

So I would suggest simply using `if dir_exists`.

turbobastii: Starting to like meson more and more. Changed it :+1:

PR #1144, opened 2022-10-27
Update white tile icons

danirabbit:
Some of these are pretty subtle, but make sure we're using the same white tile base across the set. Borders, radius, gradient, shadows, etc

For font viewer I used the image from the mimetype


micahilbery (approved):
:+1:

PR #1146, opened 2022-11-05
WIP: Actions: Make edit-clear more consistent

### LINKED ISSUE #1126: Clear icon inconsistent
### What Happened?

The clear (`edit-clear`) icons are inconsistent between normal and symbolic variants.

### Steps to Reproduce

e-icon-clear-icons

### Expected Behavior

These should probably all use the same symbol. Although I'm not sure what to do as the broom looks a little like a paint brush in symbolic form, and the backspace shape is difficult to see when used as a normal icon.

I tried just the backspace style red. This makes it easy to see in dark and light themes, and gives it the "danger/destructive" color.

e-icon-clear-1

But, I do like the broom too. Not sure what would be best here.

### OS Version

6.x (Odin)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

newhoa:
@danirabbit , you mentioned in [the issue](https://github.com/elementary/icons/issues/1126#issuecomment-1175232063)

>GitHub search and see where `edit-clear` is used and switch all of those instances to `edit-clear-all` with fallbacks enabled...

I wasn't totally sure what enabling the fallbacks meant or where exactly to look. I checked for symlinks and the like and didn't find any that needed changing, but wasn't sure what more to do. The rest I think should be on the right track. Moved `edit-clear` Broom icons to `edit-clear-all`, make `edit-clear` all use the red Backspace symbol. I did add the `error` class to the symbolic `edit-clear` icons to make them red if that is okay.

One thing I'm not sure on is the sizes. I did a new `edit-clear` based on the `mail-mark-important` icon as mentioned at 16px. Should this be done at other sizes? I'm not sure that these icons are used beyond the text-entry boxes, so I'm not sure that 32x, 48px, etc are needed (and if not should the larger `edit-clear-all` icons be removed as well?).

Current:
current

Proposed:
proposal1

Also played with making the new regular `edit-clear` a few pixels longer vs a little more square (edit-clear0), wasn't sure which to go with but I like the longer one a little more, it's a bit more legible:
proposal-square

Git Log:
- `edit-clear`: red backspace symbol
- `edit-clear-symbolic`: red backspace symbol, error class added
- `edit-clear-all` and `edit-clear-all-symbolic`: broom symbol

Fixes #1126


newhoa:
I by Github search you meant search other projects, it looks like there might be 3 places in other projects:

- [switchboard-plug-printers: dialog](https://github.com/elementary/switchboard-plug-printers/blob/ae32a14cd2c59e3346c8c7667e9921e49c3171a1/src/Dialogs/ClearQueueDialog.vala#L26)
- [installer: try install view](https://github.com/elementary/installer/blob/d47e20ad971d4ffcd20a7e38975758960e840629/src/Views/TryInstallView.vala#L47)
- [icon-browser: category view](https://github.com/elementary/iconbrowser/blob/ea167a72d36000293aa56bfdc791b33386636be8/src/CategoryView.vala#L246)

newhoa:
I'll also add, while `edit-clear` is in the freedesktop.org icon naming spec, `edit-clear-all` is not.

In places where they are used, there doesn't seem to be any actionable difference between the two -- they both clear all.

PR #1147, opened 2022-11-08
Redesign system-file-manager

### LINKED ISSUE #1101: Redesign system-file-manager
### Problem

system-file-manager

* The icon indicates `home`, but Files doesn't necessarily always open to `~/`
* The shape conflates it with a folder instead of an app
* The size is much shorter on the canvas than other app icons

### Proposal

* Avoid the `home` methaphor
* Consider a rounded square base like most other app icons

### Prior Art (Optional)

* It seems like blue is a pretty dominant color here in other file manager icons
* Including the spyglass metaphor is kind of interesting, but I'm not sure it's any more relevant than other apps that have search

## Chrome OS
chrome

## GNOME
gnome

## iOS
apple

## Windows
windows

## Other

amaze

danirabbit:
Fixes #1101


DanielTolentino:
Background tab with Search icon works better, especially to disassociate with the `home` folder as proposed.
It blends well, and it's easy to grasp because the majority of folder icons have background tabs in other systems.

micahilbery (approved):
:heart: I love it!

PR #1159, opened 2022-12-17
Update border radius of preferences-system-sharing

lenemter:
I used `applications-development` icon as a base since both icons share the same color palette.


danirabbit (approved):
This looks great, nice work!

PR #1161, opened 2022-12-18
Update border radius of application-x-firmware

lenemter:


danirabbit (approved):
This was definitely a trickier one! Nice work :tada:

PR #1169, opened 2022-12-30
FileTypes: Use fd.o `text-x-generic` for text

newhoa:
Changes `text` mimetype filename to `text-x-generic` to match FreeDesktop.org Icon Naming Specification. Same change made for the `-rtl` version.

Update the symlink names in mimes/meson.build.

---

Let me know if I need to make any other changes outside of the meson.build for symlinks or anything.

Also something I noticed while editing the meson.build file for symlinks, I'm not sure if plain names like `wordprocessing` and `text` are needed as symlinks anymore (same with stuff like `image` or `package` names/symlinks). Are these used anywhere that anyone knows of?

Thanks.


danirabbit (approved):
Looks good! Happy to make this match the spec and symlink makes sense to me.

I'm not 100% sure if those names are used, but I'd rather include the symlinks and have them unused then remove the symlinks and accidentally break something. A lot of these mime symlinks have been added over the years to fix various bugs. It might be worth going through git blame and seeing if we can add comments to the build file documenting that they're all for

Issue #985, opened 2021-01-01
Public folder symbolic icon doesn't have enough contrast in dark style


JoseExposito:
## What Happened

This icon: https://github.com/elementary/icons/blob/master/places/symbolic/folder-publicshare-symbolic.svg
Does not have enough contrast in dark style.

![Public icon bug](https://user-images.githubusercontent.com/1335948/103441845-1dbe3a00-4c51-11eb-9aff-a07b111b852d.png)

## Expected Behavior

The icon should have more contrast or the folder version should be used when importing images to photos.

## Steps to Reproduce

1. Open the Photos app
2. Click on Settings -> Import from folder
3. The public folder icon doesn't have enough contrast

## Platform Information

elementary OS 6

lainsce:
Check out #938 I think it's the icon that is used here, but adapted to be more visible in dark mode.

JoseExposito:
Yes, it could be that icon. Let me know when your PR is merged so I can test it and close this issue :+1:

PR #1181, opened 2023-03-01
Actions: Make send and public folder purple

newhoa:
This changes the send-to icons (and folder-publishare symlinks) from grayscale to purple. This better matches the idea of the icon theme that share = purple. It mirrors the purple being used by document-send and emblem-shared icons.

It also switches to the newer semi-transparent borders in the circles, which should look a bit crisper in dark gtk themes.


newhoa:
Wow, looks like this was already proposed in !938, even used the same title! What a coincidence. Not sure why it was closed.

Anyway, looks like this one closes an issue (if accepted):

Closes #985

PR #1182, opened 2023-03-06
Devices: Remove unused symlinks

newhoa:
Remove outdated/unused and non-freedesktop.org symlinks.

Removed:

- camera
- drive-cdrom
- media-cdrom
- music-player
- printer-remote
- system

There are probably more but these are the ones I've checked the codebase for and they seem not to be used anywhere. They have been deprecated for a very long time.


danirabbit (approved):
Looks good! I can verify these are not fd.o names and are unused afaict

PR #1200, opened 2023-05-17
Rename media-memory-sembolic.svg to media-memory-symbolic.svg

1280px:
I'm not entirely sure if this was done intentionally (maybe some historical typo that is kept for compatibility?) or is an actual typo, but, considering I haven't found any mentions of this icon's name outside of eOS icons' repo, I believe it's the latter.


danirabbit (approved):
Woops, that went unnoticed for a long time. Thanks!

PR #1205, opened 2023-06-16
Actions: update insert-image-symbolic

danirabbit:
## BEFORE
Screenshot from 2023-06-16 10 26 58

## AFTER
Screenshot from 2023-06-16 10 32 37


micahilbery (approved):
Love it!

PR #1206, opened 2023-06-19
Devices: Improve borders for optical media

newhoa:
Use semi-transparent borders for optical disc. This should give the icons a sharper look, especially in dark themes.

Adjust shadows to be more consistent with current shadow style of other elementary icons.


danirabbit (approved):
Solid update, nice work!

PR #1207, opened 2023-06-19
Actions: Make media-record more consistent

newhoa:
The 48px media-record icon was updated in a192e78 ([ref to media-record](https://github.com/elementary/icons/commit/a192e7847c3b1fe2a98c81befd966ac084b6f183#diff-701885c65f2e4ed3ef3eaa7f30dd44303fd31caa20e8796b1a58d509d7d7f2c3)) to be a little brighter and user the newer palette.

This updates the other sizes to give the same look. Also updates shadows and switches to semi-transparent borders for sharper look, especially in dark themes.


danirabbit (approved):
This looks really good, thanks for the attention to detail!

PR #1208, opened 2023-06-19
Actions: Update open-menu and system-run borders

newhoa:
Use semi-transparent borders on the "gear" icons to give them a sharper look, especially in dark themes.

---

I have `system-run` in 16px and 32px, but wasn't sure if they're wanted/needed. I have seen the 16px size used in other non-elementary apps. If that's something you all would like let me know and I'll update this PR. Here's the icon for reference.

system-run


danirabbit (changes requested):
Hey thanks for this branch!

In `open-menu` it looks like you're missing the highlight in the 24px here. Other sizes look way better than main.

Gear shapes are way better in 24px `system-run`! 

Adding the missing sizes would be much appreciated as well, thank you!

newhoa:
> Hey thanks for this branch!
> 
> In `open-menu` it looks like you're missing the highlight in the 24px here. Other sizes look way better than main.
> 
> Gear shapes are way better in 24px `system-run`!
> 
> Adding the missing sizes would be much appreciated as well, thank you!

Ah, I did miss that highlight! Good catch! Thanks!

Fixed that and added the other sizes. I had to force push (maybe because I amended my second commit after committing it?). Hope that doesn't cause any problems.

Let me know if anything else needs to be adjusted or anything.

Thanks!

danirabbit (approved):
These are fantastic, thank you!

PR #922, opened 2020-07-15
24px symbolic rounded arrows

danrabbit:
Possible solution to #246 

Screenshot from 2020-07-15 15 44 56@2x

PR #1209, opened 2023-06-20
Actions: Use rounded arrow for browser-download

newhoa:
Completes part of Issue #922 (Make all of the arrows consistent)

### DIFF
--- a/actions/16/browser-download.svg
+++ b/actions/16/browser-download.svg
@@ -1,15 +1,38 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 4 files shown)


newhoa:
Arrow might be a little too low on 32px, any of these better?

Current

browser-download-pr1-current-too-low

Higher arrow
browser-download-pr1-prop-higher

Higher and wider arrow (I think this one might best?)
browser-download-pr1-prop-higher-wider

Any other feedback would be appreciated. I tried to stick as close as I could to the `go-down` and `document-import` icons.

danirabbit:
@newhoa these look really good, I think you're right about the 32px the higher and wider looks best to me

newhoa:
> @newhoa these look really good, I think you're right about the 32px the higher and wider looks best to me

Ok thanks. Updated, should be good to go if everything else looks good!

danirabbit (approved):
Fantastic :tada:

PR #1213, opened 2023-07-10
Actions: Use rounded arrow for view-refresh

newhoa:
Completes part of Issue #922 (Make all of the arrows consistent)

I tried to kind of slightly round the start of the arrow line as well (and add a little highlight). Not sure if it's noticeable in use. I've been looking at them way too long lol.

---

Also, just out of curiosity, some symbolic icons are filled #666666 and some #555761. Which should I be using? I know they use the FG color when in use anyway, but just for consistency I guess. Thanks.


danirabbit (approved):
These look great, nice job! Thanks for fixing the highlight on the little tails as well :)

I've been filling with `#555761` which is Silver 900 in our palette

PR #1214, opened 2023-07-11
Actions: Use rounded arrow on some 24px icons

newhoa:
- Document Import/Export
- Object Merge
- Pane Hide/Show

---

If these need to be split into separate PRs let me know, I thought this might be okay since it was just a few. Thanks.


danirabbit (approved):
Way better, thank you!

PR #1215, opened 2023-07-12
Actions: Use rounded arrow for object-rotate

newhoa:
Also for object-merge at 16px (24px was done in last commit)


danirabbit (approved):
Much improved, thank you!

PR #1217, opened 2023-07-14
Actions: Add some semi-transparent borders

newhoa:
This adds some semi-transparent borders to a few icons that did not have them at certain sizes (mostly document actions). Should look a bit sharper with dark backgrounds.

---

Sorry, hope little things like this aren't too annoying!

Current:
16-px-borders-current

Proposed:
16-px-borders-proposed


danirabbit (approved):
Nice work!

PR #1218, opened 2023-07-14
Actions: Refresh zoom icons at 16px

newhoa:
Use semi-transparent borders, use gradient for fill, add highlight inset border, adjust rounded edges to match curves of other similar action icons. Should look a little smoother. And sharper on dark backgrounds.

---

Current:
16px-zoom-current

Proposed:
16px-zoom-prop


waldyrious:
IMHO the current version has better contrast; not sure the highlight is a net positive. No opinion on the other changes, though I welcome improved sharpness in dark mode :)

danirabbit (changes requested):
It looks like we lost quite a bit of border radius here compared to other tile icons at this size. I think I would base these on categories/16/applications-other to make sure we're being consistent there

While we're here, I wonder if we should change these glyphs to be more contrasty as well like:

Screenshot from 2023-07-17 10 36 10

micahilbery:
> It looks like we lost quite a bit of border radius here compared to other tile icons at this size. I think I would base these on categories/16/applications-other to make sure we're being consistent there
> 
> While we're here, I wonder if we should change these glyphs to be more contrasty as well like:
> 
> Screenshot from 2023-07-17 10 36 10

The border is a big improvement, but I agree. I think they are too pointy now and I also agree that they should be more contrasty.

PR #1224, opened 2023-07-26
Devices: clean up 128 drive-harddisk

danirabbit:
This icon still needs a much bigger redraw, but hopefully this makes it suck ever so slightly less for now


newhoa:
I tried a little refresh to the solid state icon a few months ago but got side tracked and never finished. Tried to update the symmetry, with updated borders/gradients, and rounded edges.

drive-harddisk-solidstate   solidstate-prop2  

Right direction? Something I should continue working on? Figured I'd throw it here since it fits the subject (or if anyone else wants to pick it up). Any feedback would be appreciated!

micahilbery (approved):
LGTM 👍 I think if nothing else the borders/strokes are a big improvement.

Issue #1168, opened 2022-12-24
document-save should look less like a mimetype


newhoa:
### Problem

`document-save` and `document-save-as` look like mimetype icons. `document-save` is a symlink to `text-x-install`. Also because of the coloring, they can be confused with disabled icons.

Current:

![document-save-current](https://user-images.githubusercontent.com/1984060/209435254-7f97d397-da8b-4f9e-bb1a-3f6408e53021.png)

### Proposal

It would be nice if this looked a little more like an action or button style icon.

newhoa:
Just quickly tried out the Download emblem. I like it, but I wonder if using it for an emblem, folder, and save icon would be too overlapping. Maybe there could be some variation for it (wonder if a mix of that and the "partial download" arrow might work)?

![document-save-prop1](https://user-images.githubusercontent.com/1984060/209433933-20b2dc60-676d-4f97-afba-34775ac3e903.png)

newhoa:
Here's another proposal, more closely matching the browser-download, and document import/export icons:

![document-save-prop2](https://user-images.githubusercontent.com/1984060/210552873-71bd50b5-2ea5-4b47-b968-44b36650ab52.png)

prop2 was kind of meant to represent a "save changes to the document" look, but it looks a little close the import icon? Not sure about that one.

prop1 looks a little more right maybe? More of a hybrid of download and export. And not too far off from the current document-save, just less mimetype-looking using the "document" symbol of import/export.

Still would need to make a "save as" version.

Any thoughts or suggestions would be helpful!

fitojb:
The Save action doesn’t need wheel-reinventing because someone feared TikTok kids were dumb. The tried-and-true diskette metaphor is still the most usable, and that hasn’t demonstrably changed across generation.

danirabbit:
@newhoa I would be down with making it more like the other document icons using a tile shape as the base and using the same color arrows as the download/import/export icons. I think that reasoning makes sense.

For "Save as", what about using a folder icon as the base? Since this opens the file chooser that might make more sense than trying to convey file renaming

1280px:
I think Firefox icons might be a useful reference too:


![eos_load](https://github.com/elementary/icons/assets/71165491/a69f049a-3d5f-4ecf-854a-53fbd071f8ef)
Open file = "take" a file from a directory (up arrow)

![eos_save](https://github.com/elementary/icons/assets/71165491/ab175eee-ebf9-4dd7-bd4a-f0a25ddf0e78)
[...]

PR #1230, opened 2023-09-10
Actions: New document-save icon

newhoa:
Use same document symbol that is used in document import/expot.

Addresses part of #1168


danirabbit (approved):
Totally on board with this, this looks great. Thanks!

PR #1233, opened 2023-09-20
Actions: Horizontally center media-playback-start

### LINKED ISSUE #1232: Icon inside the play button not in center
### What Happened?

The play icon within the button is slightly to the left. It is not causing any problems, just to improve the aesthetic.

image

### Steps to Reproduce

* Play music
* Click on the sound button in the wingpanel
* Pause the music so that the play button is visible

### Expected Behavior

Icon in the center of the button

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #1232


jeremypw (approved):
Looks pretty central to me when installed. Not sure if the best center is the geometrical one or just half the width but the difference is probably not visually significant at these sizes.

PR #1234, opened 2023-10-29
Categories: Use bell for notification settings

newhoa:
This is an alternative proposal to PR #803. Uses the same shape as the symbolic notification icon and notification-disabled.

Screenshot_2023-10-28_20-06-06


danirabbit (approved):
I think I'd like to follow this up with a branch that adds a bit more depth, but this is better than what's in main imo! Thanks :)

PR #1238, opened 2024-01-29
Add Rust icons based on the official raster versions

aerkiaga:
Instead of using the SVG artwork as in #983, here I attempt a different method, namely tracing over the PNG images and exporting the result as SVG.


danirabbit:
Hey thanks for your contribution! I just responded to the original issue report. Sorry I know it's not what you're hoping to hear!

>
>

PR #1239, opened 2024-01-29
Status: halign checkmark on 32px process-completed

danirabbit:
I feel dumb for making this pull request but I think it looks more centered half a pixel to the left 😆️


micahilbery (approved):
No, you're right. And now I can't un-see the old one as off-center. 😆👍

PR #1246, opened 2024-02-23
Round window control icons

danirabbit:


micahilbery (approved):
:heart: Love it! Cuter, softer, and friendlier in general. Kinda wild how a fairly minimal visual change can really change the tone.

Issue #1240, opened 2024-02-01
usb-receiver


danirabbit:
### Problem

Missing icon name `usb-receiver`:

![Screenshot from 2024-01-31 16 08 15](https://github.com/elementary/icons/assets/7277719/965638fb-9098-4f22-8a40-a6de56cdccab)

### Proposal

This is for the wireless keyboard/mouse receiver

newhoa:
Closed/Fixed with #1247

PR #1247, opened 2024-03-01
Usb reciever

wpkelso:
This adds icons for the USB universal mouse/keyboard receiver, closing #1240 . The set uses `drive-removeable-media-usb` as a base.


danirabbit (approved):
Hey this is really good! Thank you! :tada:

Issue #1065, opened 2021-12-26
Make more use of color in cursors


danirabbit:
### Problem

Colors are historically very monochrome, but we could be a lot more expressive with color

### Proposal

Find ways to integrate color into cursors, maybe even accent colors?

### Prior Art (Optional)

https://github.com/ful1e5/Bibata_Cursor

input-variable:
Hey! First-time contributor here, decided to dip my toes in the water and take a crack at this. 

![image](https://user-images.githubusercontent.com/11876024/147427743-46830c9b-d131-4f3f-b2b9-14fc1f1e4799.png)

Straying away from the proposal a bit, I focused on applying accent colours to the cursors. First, I tried applying a gradient fill on the left pointer (accent 700 -> 500). They look okay but I'm not too sure how they'll behave in terms of contrast/accessibility so further testing is needed.

I also mirrored the 700 -> 500 gradient on some of the other cursors. For the pencil cursor, I didn't apply a gradient for the base colour (keeping it in line with the original). Instead, I opted for the 700 accent variant over the 500 as the highlight added some brightness.

The shadows might be a good place to add an accent colour but once again, I'm not too sure how that might play out contrast-wise. 

![image](https://user-images.githubusercontent.com/11876024/147429721-985d668a-00ae-420b-a925-26d0f5292537.png)

I've also noticed that if accent colours are used for the modifier then a problem arises with symbols such as "copy" and "no-drop". By using the same accent colour for the fill of the small circle, we lose the effect of the original colour but in return gain a more visually cohesive style.  In the image above, I've presented four different combinations of the "copy" cursor in three different colours. The text above shows the fill of the circle for the entire column.

![image](https://user-images.githubusercontent.com/11876024/147429279-4c909282-2732-4334-ac9d-af6f771023e3.png)

Going back to the proposal, I added some more features to the pencil cursor and gave it colours that represented its real-world counterpart. More abstract cursors like the arrows/grab could potentially share one colour in a set (see below).

![image](https://user-images.githubusercontent.com/11876024/147430368-8c782a8d-4577-4916-9ae9-73629268730c.png)

If you have any questions/feedback, let me know. Hope this could help!

danrabbit:
Hey @nakotami thanks for exploring this! I think I'm more in favor of the direction of using consistent colors for like cursor families or modifiers or things like what you've done with the pencil icon, rather than focusing on recoloring the cursor body itself.

[...]

PR #1248, opened 2024-03-02
Adding color to cursors

wpkelso:
This does some basic work towards addressing #1065 , starting with the `watch` animation in both its standalone and `ptr` form. I liked nakotami's idea with the `pencil` coloring from [here](https://github.com/elementary/icons/issues/1065#issuecomment-1001309184), so that has been added in as a starting point; other "real" cursors will eventually follow suit (`zoom-in`/`zoom-out`).

--EDIT--

This PR now encompasses not just the work addressing #1065, but also small modifications to many other cursor files to bring them more in-line with the rest of the set in addition to clarity improvements.


wpkelso:
Current status of things

I tried not to touch the main bodies of cursors where possible, as well as keep similar icons on similar color-lines. I think it might be a good idea to keep the all-black set of cursors as an option available to the user for accessibility purposes.

danirabbit (changes requested):
What do you think about putting all the symbols for dnd-cursors on a colored circle like `dnd-copy` is? Some of these symbols are a bit harder to see in these versions. For example this feels like a clear step back for legibility:

Screenshot from 2024-03-21 11 55 13

I'm a little worried with making the working cursors blue that folks will naturally think this should be following their system accent color. Maybe there's something interesting we can do with rainbow colors here? Or maybe revert this for now and revisit

It looks like you accidentally included some changes for USB receiver icons in this branch

wpkelso:
Not sure what happened with the USB changes, nothing actually changed there to my knowledge. However, I've reverted that file to main.

I reverted the watch cursors as well, I'll revisit them in the issue thread for a more thorough design exploration.

Changing to symbol-on-circle definitely helps with clarity for the dnd-cursors, although the size constraints meant I had to do some reworking of the symbols themselves.

danirabbit (changes requested):
If the waiting cursors were their own branch, I would merge those today. I think that's an easy win. Those look great

For some of the other icons, I think there's more work to do here. Especially making sure we're consistent with the main icon set, using the colors and icons and styles already established there. I'll open a PR today with what I mean there

I'm not really fully convinced on the way the magnifier icons have been colored. I think generally updating separate families in separate branches might be better here so we can merge in what's working and consider/iterate on others separately

There's also still some issue with this PR touching the USB receiver icon, which it shouldn't

wpkelso:
@danirabbit I think the original badges I was working with here were based on an earlier version of the HIG versus what you've done in #1250 , and I just failed to realize it.

I'll go ahead and split this up and close this PR. Hopefully that will resolve the issue with the USB receiver icon (I thought it had been unified with what was on main but perhaps that got pulled in before I had added the precommit on my end and thus wasn't cleaned up on main).

PR #1249, opened 2024-03-28
Add applets-screenshooter

danirabbit:
This is not an fd.o icon name but portals, for example, expects it to exist


danirabbit:
I opened https://gitlab.freedesktop.org/xdg/default-icon-theme/-/merge_requests/4 to propose standardizing the screenshot app icon name

PR #1250, opened 2024-04-19
Redesign dnd badges

danirabbit:
Makes sure cursor badges:
- [x] are the same size
- [x] in the same position
- [x] Use palette colors
- [x] Use the same metaphors as the main icon set

`dnd-link` is actually about creating a symbolic link, not a hyperlink. So this uses the same curved arrow metaphor as in emblems


danirabbit:
I'm not sure what the `link` icon is actually for. It's not the icon that is used when a hyperlink is hovered and it's not the dnd symlink/alias icon. I can't find any demos that use it. No idea where that comes from. So since I can't determine its purpose I didn't touch it

wpkelso:
Is there a reason to use both a blurred drop shadow and a constructed drop shadow in these? From my understanding of the system, the blur tends to be used to define exterior shadows (i.e. icon vs background) while the constructed shadow with multiple circles seems to be used more for definition between shapes within the icon; here, they're used in the same context of defining a shape against the background.

PR #1251, opened 2024-04-19
Redesign watch to accommodate color

wpkelso:
Split of #1248 focusing on redesigning `watch` to better accommodate color.


danirabbit (approved):
Love these, great work!

PR #1252, opened 2024-04-19
Cursors: Redesign and color zooms

wpkelso:
Redesigns the zoom cursors to allow for more color and shading, and increases the material expression. I tried to match the size and placement of the original's lens, but thickened the handle quite a bit to get more room to work.


newhoa:
Going by the advice in the [previous conversation thread](https://github.com/elementary/icons/pull/1248#pullrequestreview-2011884625), trying to use aspects of the main theme might still apply here. If so, maybe using the `edit-find` magnifying glass would work in this case? Also, the zoom icons are the rounded-squares, and going by the [advice here](https://github.com/elementary/icons/pull/1218#issuecomment-1638741337) could use some better contrast.

So combining the two ideas above ends up with something like this:
cursor-zoom-prop


SVGs:
zoom-in zoom-out

---

I do find the main `edit-zoom` icon a bit difficult to see on dark backgrounds, though.

For the elementary-xfce set we came up with this one:
edit-find

And combined with some of the other ideas:
cursor-zoom-prop2

But not sure if that would be a welcomed change for the main theme or work well here.

---

Anyway, don't mean to step on any progress here, just wanted to throw some ideas out and add to the conversation. Love the work you've been doing @wpkelso!

danirabbit:
@newhoa I love these! Please propose the magnifying glasses for the main icon theme too they're great

wpkelso:
I think the first two ideas in the last screenshot (with the plus/minus interior to the lense) are quite nice looking, and a little more legible than using the rouded-square `zoom` icons, even if it's more consistent with elsewhere.

---

Also, some rough explorations of the xfce color scheme applied to this PR's proposal:

image

image

I think it's still too disconnected from everything else to really work well for the primary scheme, but I figured it was a direction worth exploring at the least

wpkelso:
Somehow managed to close this while fixing some local stuff. Re-opening with some options derived from the `edit-find` rework

image

And at 1:1

image

PR #1253, opened 2024-04-22
Actions: Use brighter colors for Find icon

newhoa:
Ok, submitting a PR for making the `edit-find` magnifying glass a little brighter (as mentioned in #1252).

Making a draft since I want to throw out some options first.

First, the elementary-xfce icon:
V2-light-xfce-prop V2-dark-xfce-prop

It's a little different in design and proportions than the current icon (and a bit smaller as it was made to provide the 2px margins that some other action icons have).

Current icon:
current-light current-dark

I wondered what the same idea combined with the current icon would look like:
V2-light-prop
V2-dark-prop

Top row is simply the current elementary `edit-find` brightened up. Second row is the current elementary icon combined with the colors from elementary-xfce.

On these two rows, the handles have slightly different colors and borders. Not sure if it makes much of a difference, but if there's a preference between them just let me know which looks better.

So...
1) Go with elementary-xfce icon?
2) Go with current elementary icon brightened?
3) Go with current elementary icon with elementary-xfce colors?


danirabbit:
I think I like the shape of the current elementary icon a bit better, especially the handle to me seems more defined, and I like the proportions a bit more. But the colors from the XFCE one are way better, especially on dark backgrounds. So, I'm glad you did that bottom row because they look fantastic.

Haven't looked too closely at the differences between each option on the bottom row, but if you wanted to expand on what the differences are/why etc happy to provide feedback.

Thanks for opening this!

newhoa:
Ok, I think these should be ready for review!

I replaced the old 16px icon. If there was some reason for keeping the monochromatic-ish icon let me know and I can change it back.

Also, I did adjust the handle alignment (it was slightly off-center, down and to the left). If the off-center handle works better I can change it back, but I thought the centered handle looked a little better in use.

> Haven't looked too closely at the differences between each option on the bottom row, but if you wanted to expand on what the differences are/why etc happy to provide feedback.

It's okay, probably not worth the time. Coming back to it the only one that I can tell a difference on is the handle on L is slightly darker. For what it's worth I went with K.

Thanks!

danirabbit (approved):
These are great, thank you! Also thanks for adding the 16px find-replace and fixing 16px find

PR #1256, opened 2024-04-22
Cursors: rename to fit spec

### LINKED ISSUE #895: Mouse Cursor in FIREFOX fault
Elementary OS 5.0 Juno. 

Unable to get a screen shot with the mouse cursor in. 
But, when you hover over a link in firefox rather than the nice black hand and finger cursor being displayed you get this old windows 95 looking sideways one instead. Only noticing it in Firefox so far. 

Nothing major, but is consistent over several fresh installs of JUNO :) 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/91377806-mouse-cursor-in-firefox-fault?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F27377189&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1073: Xcursor library is unable to find cursors that are unique to theme Adwaita, which have no equivalent in the default theme (elementary)
### What Happened?

Issue: The [Xcursor](https://www.x.org/releases/current/doc/man/man3/Xcursor.3.xhtml) library is unable to find cursors that are unique to theme Adwaita, which have no equivalent in the default theme (elementary). 

Here is the list of cursors unique to Adwaita, with no equivalent in theme elementary.
```
$ pwd
/usr/share/icons

$ comm -23 <(ls Adwaita/cursors) <(ls elementary/cursors)
alias
col-resize
context-menu
default
dnd-no-drop
e29285e634086352946a0e7090d73106
e-resize
ew-resize
ne-resize
nesw-resize
n-resize
ns-resize
nw-resize
nwse-resize
pointer-move
progress
row-resize
se-resize
size_all
size_bdiag
size_fdiag
size_hor
size_ver
s-resize
sw-resize
wait
w-resize
```

Since cursors such as "col-resize" (which is unique to Adwaita) are being used for window resize across the OS, I presume there is already a fallback mechanism in place. However for Xcursor to fallback to Adwaita, the `index.theme` file on theme elementary needs to be updated with the `Inherits=Adwaita` entry. 

Xcursor manpage describes the search process is detail as follows.

```
$ man xcursor

--snip--
   THEMES
       Xcursor (mostly) follows the freedesktop.org spec for theming icons.  The default search  path
       it  uses is ~/.icons, /usr/share/icons, /usr/share/pixmaps.  Within each of these directories,
       it searches for a directory using the theme name.  Within the theme directory,  it  looks  for
       cursor  files  in  the 'cursors' subdirectory.  It uses the first cursor file found along  the
       path.

       If necessary, Xcursor also looks for a "index.theme" file in each theme directory to find  in‐
       herited themes and searches along the path for those themes as well.

       If  no theme is set, or if no cursor is found for the specified theme, Xcursor checks the "de‐
       fault" theme.
--snip--
```

### Steps to Reproduce

The following function call will not find cursors such as "col-resize" which are unique to theme Adwaita.

```
Cursor
XcursorLibraryLoadCursor (Display *dpy, const char *file);
```

But will work after updating the `[Icon Theme]` section of `/usr/share/icons/elementary/index.theme` with `Inherits=Adwaita` like follows

```
$ head -8 index.theme 
[Icon Theme]
Name=elementary
Comment=Smooth modern theme designed to be intuitive.

Example=directory-x-normal

Inherits=Adwaita
```

### Expected Behavior

The following Xcursor API should be able to find and load cursors that are unique to theme Adwaita. For example cursor "col-resize"

```
Cursor
XcursorLibraryLoadCursor (Display *dpy, const char *file);
```

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #1255: Only support the GTK/CSS cursor names
As per https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/commit/74e9b79471236320d2af4925d6c5bb7df22380ce we should only ship the GTK/CSS cursors as Mutter will only use them

danirabbit:
Closes #895
Fixes #1255
Fixes #1073

* Drops cursor names that are no longer supported/aren't in the CSS spec
* Renames existing cursors to spec names
* Make sure we have symlinks for cursors we aren't covering right now
* Adds context menu cursor
* uses a colorful badged help icon instead of the large question mark


tintou (approved):
this looks good to me!

PR #1258, opened 2024-04-22
Cursors: new base cursor design

danirabbit:
Based on #1256 

* Redraw base arrow to have rounded corners and thicker stem
* Make the stemless arrow use the same arrow body
* Use same shadow technique as main icons to prevent shadow bleed and make sure shadows are consistent even with different shapes


danirabbit:
@wpkelso would love your feedback here as well :)

cursors/24/no-drop.svg
wpkelso, line 1: From the preview window, this symbol feels slightly harder to read than it was. I think the background sandwiched between the two white circles might be messing with spatial perception a bit and muddying it

cursors/24/default.svg
wpkelso, line 1: Definitely like this shape, its quite playful

cursors/24/help.svg
wpkelso, line 1: It seems like the outline might be interfering with the question mark at smaller sizes. The dot in particular feels like it disappears into the outline

wpkelso (commented):
The white outlines on the emblems feel to me like they're interfering with the white symbols a fair amount. Perhaps some middle ground with a flatter emblem and a darker outline would work better in this respect?

cursors/24/help.svg
danirabbit: Yeah valid, here's a few variations trying some different things. Definitely adding a bit of shadow on the symbol helps to define the shape a bit imo. And then there's a variation without a solid color and a variation with a darker outline

Screenshot from 2024-04-22 15 33 24

danirabbit: I have another set of variants here with trying to add a bit of dimension to the badges:

Screenshot from 2024-04-22 16 12 23

wpkelso: I think column 3 separates the elements the best

cursors/24/no-drop.svg
danirabbit: Do you feel like this is resolved now?

wpkelso: Yes, much easier to read than it was

cursors/24/progress_01.svg
wpkelso, line 1: I wonder if a mesh gradient would work here for smoother transitions? I don't know whether rsvg supports them, but it could be a thought for the future.

wpkelso (approved):
I like how consistent everything feels now

danirabbit:
Thank you so much for your help!

PR #1271, opened 2024-05-03
Devices: Make input-mouse more generic

wpkelso:
Should close #881

Main effort was towards making the specific mouse more generic, rather than emulating the Magic Mouse, as well as pushing towards more parity with the symbolic icon---which does have buttons.

Issue #881, opened 2020-04-21
Update mouse icon to look like most mice


philipzae:
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**
Yes the mouse icon isn't easily understood at small sizes found in menus and doesn't like most mice used by the majority of people.

**Describe the solution you'd like**
Designing a new icon or utilizing one from another theme

**Existing work**
The humanity icon theme has a design that could be used.
![preview](https://imgur.com/FD3bN11.png)

**Describe alternatives you've considered**
Switching the icon theme to something else

**Additional context**
I've filed this bug downstream with Xubuntu's default icon theme as well.
https://github.com/shimmerproject/elementary-xfce/issues/214

alfonsojon:
I believe the current mouse icon falsely implies likeness to the Apple Magic Mouse, which is an uncommon mouse that most users are not familiar with. I do think there should be a better mouse icon, one that is simple and easy for all users to understand without association to a certain brand.

Fatih20:
I agree. Are you going to try to design one? If not, I'll be looking into it.

philipzae:
> I agree. Are you going to try to design one? If not, I'll be looking into it.

Please go ahead and do so, while we await for the elementary team.

marcosins:
(came here from the downstream issue)

I would like to add some references:

### Adwaita
16x16 | 48x48
------------ | -------------
![Mouse](https://raw.githubusercontent.com/GNOME/adwaita-icon-theme/mainline/Adwaita/16x16/legacy/input-mouse.png) | ![Mouse](https://raw.githubusercontent.com/GNOME/adwaita-icon-theme/mainline/Adwaita/48x48/legacy/input-mouse.png)
![Touchpad](https://raw.githubusercontent.com/GNOME/adwaita-icon-theme/mainline/Adwaita/16x16/legacy/input-touchpad.png) | ![Touchpad](https://raw.githubusercontent.com/GNOME/adwaita-icon-theme/mainline/Adwaita/48x48/legacy/input-touchpad.png)

### Yaru
16x16 | 48x48
------------ | -------------
![Mouse](https://raw.githubusercontent.com/ubuntu/yaru/master/icons/Suru/scalable/devices/input-mouse-symbolic.svg) | ![Mouse](https://raw.githubusercontent.com/ubuntu/yaru/master/icons/Suru/48x48/devices/input-mouse.png)
[...]

PR #1272, opened 2024-05-05
Devices: Add buttons to touchpad to match symbolic

wpkelso:
Related: #881 , #1271 

Adds left and right buttons to the touchpad to increase clarity out of context and to more closely match the symbolic version.


danirabbit:
I don't love this change, especially since it's more common for touchpads on laptops to not have external buttons. I think it might be more worth looking at where/if this symbolic icon is even used (since this is outside of the fd.o spec) and maybe either dropping the symbolic or considering how it can be adjusted in context.

PR #1281, opened 2024-05-17
Actions: Refresh revert icon

newhoa:
Use new document shape, rounded arrow, newer style gradients. Make red to indicate risky action and differentiate it from the undo icon.

I'll throw some other options at the bottom, let me know which one is liked best and if this style is okay to go with and I will finish the other sizes. Thanks!

Current:
document-revert-current

Proposed:
document-revert

prop1

Other options:

Keep yellow arrow, use new document shape:
document-revert-yellow

Keep document shape, use red arrow:
document-revert-doc

Tried keeping the document shape on that last one, wasn't sure if overusing the square document would be too confusing with import, export, save, and revert.

### DIFF
--- a/actions/24/document-revert.svg
+++ b/actions/24/document-revert.svg
@@ -1,267 +1,289 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
-   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-   xmlns:svg="http://www.w3.org/2000/svg"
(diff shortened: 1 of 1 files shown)


newhoa:
Updated this one, went with a deeper red (and added 16px, -rtl icons).

prop2

The 16px one looked weird without raising the arrow a bit. Raising the arrow made it look clearer.

16px-prop1

If the 24px one needs to be changed to match (or vice versa) let me know). It would look like this:

prop3

Anyway, I think this should be ready for review/feedback. Thanks!

danirabbit:
Hey, so the difference between the tile shape and the portrait rectangle shape is that the tile shape represents the application, not the document. And it was yellow to match "undo" so that the metaphor would be that you're undoing all changes to the document instead of just undoing the last action.

I think if we're updating the arrow shape here, we should also make sure to match the arrow tail shape to the new undo/redo arrows

PR #1282, opened 2024-05-17
Actions: Update print preview magnifying glass

newhoa:
Use the refreshed magnifying glass from edit-find.


danirabbit (approved):
Wow way more readable! Great improvement

PR #1283, opened 2024-05-18
Actions: Refresh and add color to session icons

newhoa:
We did these in elementary-xfce, was wondering if you all would be interested in them?

The dark icons were a little close to disabled-button looking and adding color helps with that and makes them easier to see on dark backgrounds. Also reboot uses rounded arrow.

Questions:

1) Only 32px and 48px exist, do you all want 24px and 16px added?
2) ~`system-restart` and `system-suspend` are non-FDO names, remove these or think about it in another PR?~ Nevermind, after checking the code base these are uses in elementary.
3) We also did suspend and hibernate, and switch user if you all want them (none are FDO, but are commonly used) (also, suspend was a different design than current, didn't want to change it without asking):

session-prop

### DIFF
--- a/actions/32/system-lock-screen.svg
+++ b/actions/32/system-lock-screen.svg
@@ -1,99 +1,120 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
-   xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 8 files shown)


danirabbit (approved):
These are great, much nicer looking!

danirabbit:
@newhoa I would totally merge sleep and hibernate ones. The snowflake looks great and I love the color there.

We're using a moon metaphor in quick-settings also, but we're using stars instead of "z"s since that seems like a locale specific thing:

Screenshot from 2024-05-28 11 56 04

Issue #1151, opened 2022-11-10
system-software-install


danirabbit:
This is currently using the AppCenter icon, but we should make sure we're shipping the branded AppCenter icon with AppCenter and use a generic software install app icon here. This could be used for installing system packages/extensions that aren't related to appcenter like codecs or langpacks or something else

wpkelso:
Would riffing off the building idea from AppCenter be appropriate here, or is that more core to AppCenter's branding and thus should be kept separate in this case? I could see a warehouse or a library working nicely alongside AppCenter, but I'm afraid it might not generalize well.

danirabbit:
Another possibly interesting one is like a delivery truck being the thing that brings you packages

wpkelso:
Unless I'm misunderstanding the use case of `system-software-update`, I think the two ideas could work in tandem for this and #1150 , where you have some warehouse-type thing to initially give packages, then a truck to give updates. That does have the drawback of ignoring existing metaphors, but perhaps it's fine in this case to differentiate between a status and an app

danirabbit:
@wpkelso I wouldn't worry too hard about `system-software-update` since I think these days in most desktop environments updates aren't a separate app. This is probably something that could be proposed for deprecation in the naming spec imo

wpkelso:
Closed by #1284

PR #1284, opened 2024-05-29
Apps: Redesign system-software-install

wpkelso:
Related #1151 

Redesign `system-software-install` to separate it from AppCenter branding


micahilbery:
That little truck is 👨‍🍳👌💋

wpkelso:
I went ahead and did all of the sizes to keep consistency with the other apps, even if they're unlikely to be used

danirabbit (changes requested):
Hey sorry for taking a while to follow up here. This is a super cute little icon. I love that it shows a kind of more "industrial" implication of the package manager versus the app store as well.

My only major complaint is that its much smaller than other icons at its size and where the baseline of the icon is seems to be higher. See for example here:

Screenshot from 2024-06-24 08 14 57

Another thing to watch out for is how highlight gradients fall over the top edges of objects. Look at this for example. You can see how the gradient stop should match up with where that corner stop is:

Screenshot from 2024-06-24 08 18 05

wpkelso:
Oh oops, apparently 'temporary' means permanent. Let me fix that then it will be ready for review again

wpkelso:
Actually ready for re-review now

danirabbit (changes requested):
Looking really good! The smallest sizes are especially cute. Just a few comments here

The border color here is significantly different from the archive file type icon:

Screenshot from 2024-08-16 15 33 06

when you have a rounded shape that contains another rounded shape, make sure that you adjust the radii so that there's consistent spacing around the edges. See the red circle in the first shape and how there's a lot of extra gap compared to the shape on the right?

Screenshot from 2024-08-16 15 37 43

Also for highlight gradients you want the outer stops to be contained inside the stroke and the inner stops should line up with the radius handle of the shape basically

danirabbit (approved):
Thanks for sticking with it through all the rounds of feedback! These are very cute icons, good job :)

PR #1285, opened 2024-06-01
Actions: Apply new save-as design to symbolic icon

newhoa:
Hopefully this retains the "folder" look and is distinguishable enough from the symbolic "save" icon.

Let me know if anything can be done to make it better.

Thanks!


danirabbit (approved):
We probably should revisit the symbolic folder icon in the future to make this kind of thing more clear, but this works for now :)

PR #1287, opened 2024-06-06
Status: Allow finer display of battery level

wpkelso:
Related elementary/wingpanel-indicator-power#257 

New level names are derived from the naming scheme present in Adwaita


danirabbit:
`battery-caution` and `battery-low` are the only icons in the fd.o naming spec

If we want to have more fine-grained battery levels we should probably do that in `wingpanel-indicator-power` since that's [the only place battery level icons are used](https://github.com/search?q=org%3Aelementary%20battery-full&type=code)

PR #1288, opened 2024-06-06
Create prefs-system-symbolic

danirabbit:
Can't decide if this is better or worse:

Screenshot from 2024-06-06 13 35 06
Screenshot from 2024-06-06 13 36 06


wpkelso:
The less complex option feels like it fits with the rest of the row better in both it's level of detail and visual weight. That said I think I prefer the complex option in a vacuum

PR #1289, opened 2024-06-09
Actions: Colorize font formatting icons at 16px

newhoa:
Colorize font formatting icons for better visibility under dark themes. Use semi-transparent borders,
Update font to Inter-V.

Add missing FD.o icons `format-text-direction-*`

---

Initial proposal on these, feedback would be appreciated.

Questions

1) Should this also be done for the existing 24px icons, or are they even needed? (I really see almost no places where these are used anyway)
2) Would it be better to use **B** _I_ <ins>U</ins> and ~~S~~, which have become more common in editors, chat programs, etc.

Thanks.


micahilbery:
I think it would be better to use **B**, <ins>U</ins>, and <del>S</del>. I think the `a` (**a**, <ins>a</ins>, <del>a</del>) alone is harder to understand for english speakers where those things start with those letters and no more clear for non-english speakers. Also the consistency with the symbolics is a good thing as well.

kurisubrooks:
The current black/grey coloured icons are difficult to see when using a dark mode/theme in apps like LibreOffice, so the blue colour of these icons in this PR are a great change and will definitely improve visual clarity.
dark mode screenshot of bold/italic/underline/strikethrough buttons in LibreOffice Writer 26.2

I also agree with the above comment from @micahilbery, where there's a good opportunity to further improve clarity in the meaning of the icons by representing **B** (Bold), <ins>U</ins> (Underline) and ~S~ (Strikethrough) as their lettered counterparts, but this may be out of scope for this PR.

^ For reference, LibreOffice's implementation of the elementary icon set also uses this system of representation.
light mode screenshot of bold/italic/underline/strikethrough buttons in LibreOffice Writer 26.2

PR #1291, opened 2024-06-21
Categories: preferences-desktop-locale-symbolic redesign

danirabbit:
Match corner radius of other modern symbolic tile icons. Make it look a bit more like the color version by having the hiragana tile be outline style

## AFTER
Screenshot from 2024-06-21 15 19 46

## BEFORE
Screenshot from 2024-06-21 15 19 12


micahilbery (approved):
Nice! Looks great!

PR #1294, opened 2024-07-26
Align `go-home` with other `go-` icons

### LINKED ISSUE #1094: Home button lower than others
### What Happened?

The home button isn't aligned well with the other icons. 

### Steps to Reproduce

Screenshot from 2022-03-02 15 16 06

### Expected Behavior

That the home button would be well aligned than the other buttons. It's lower than the other buttons. 

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

wpkelso:
Closes #1094 

Refreshes `go-home` to use standard fill color, unifies it's corner radii with other `go-` icons, and move it's baseline up by two pixels to align with other `go-` icons


danirabbit:
It looks like this has conflicts now. Can you resolve those please? :)

danirabbit:
It also looks like the icon in #1094 is the 24px version, so you'll need to make sure to update that one as well

danirabbit (changes requested):
Couple problems here with like positioning and size. If you look at for example the home icon in an entry at 16px, it's really crammed up against the top edge now. It doesn't look vertically centered:

Screenshot from 2024-08-16 14 58 35

In terms of scale, if you were to take the 24px version and scale it down at the same time as another icon it's commonly next to (like the arrow) you can see that this 16px version is a lot larger than the arrow:

Screenshot from 2024-08-16 15 04 08

Pictured here in red is how large the home icon would be if it was scaled proportionally to the arrow icon. You can really see also how the length of the chimney is significantly different between these sizes

And also it looks like you haven't merged down these shapes :)

wpkelso:
> Couple problems here with like positioning and size. If you look at for example the home icon in an entry at 16px, it's really crammed up against the top edge now. It doesn't look vertically centered:
> 

Hmm, that might be an artifact of misidentifying the problem in the original screenshot. I believe (Ill check a little later) that 'view-refresh' is top aligned when I looked at the file, so I matched that alignment here. Centering home will likely reintroduce the alignment issue present in the screenshot without some compensation in refresh.

Issue #928, opened 2020-07-16
Improve close window icon


igaldino:
## Prerequisites
- [X] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**

I think the close window icon is too small for Elementary window headerbar. Other icons in the window headerbar are larger and looks a little bit weird. I don't see that problem happening with maximize button though.
I notice that the close icon has the same size that GNOME's. However, Elementary's headerbar uses buttons without borders, which is different from GNOME approach. The latter, because of the borders, seems a different button or action, so looks consistent, while the former looks like a papercut in the icon design.

* Elementary
![2020-07-16_16-50](https://user-images.githubusercontent.com/5329570/87718249-c94cfe00-c787-11ea-90e1-eecae18e7730.png) ![2020-07-16_16-51](https://user-images.githubusercontent.com/5329570/87718268-cf42df00-c787-11ea-9c01-89983371d7be.png)
* GNOME
![2020-07-16_16-46](https://user-images.githubusercontent.com/5329570/87718408-0fa25d00-c788-11ea-922f-c660449ab93f.png)

**Describe the solution you'd like**

I'm not a designer, so I might not have the knowledge to suggest a proper solution. What I could suggest are two options:
1. Increase the icon size. Make it the same size of the other icons in the headerbar. I think the size of window buttons vs window title size in Windows is a good ratio example:
![2020-07-16_16-52](https://user-images.githubusercontent.com/5329570/87719482-cbb05780-c789-11ea-88fa-5e8d51be3c2a.png)
2. Make other buttons in headerbar to have borders like GNOME's. Not sure if this will work because it'll add more constraints to what can fit in the button borders (check example above).

**Existing work**

I didn't find anything/anybody working on this in Elementary.

**Describe alternatives you've considered**

Just use what we have today :).

**Additional context**

No, I'm good, thanks.

matfantinel:
I can confirm that I've seen people that were new to eOS struggle a bit to find where the close icon was. Specially if the first time they try it is in an app with a busier headerbar like Mail or Files.

The placement is consistent everywhere, but first discovery is a bit hard if you don't know where to look.

Windows handles this with bigger icons (although consistency is a mess) and MacOS handles it with the colorful buttons, but I don't know if the colors would fit well into eOS

PR #1295, opened 2024-07-30
WIP: Unify symbolic window actions

wpkelso:
Mostly exploratory right now (rtl not included), but this intends to address #928 by unifying the size and language of the window controls.

Previously, `window-close` used a baseline of 4px, `window-minimize` used a border of 3px, and `window-maximize`/`window-restore` both used 4px borders. This mostly unifies to 2px border, with `window-close` now using a 3px border because 2px felt slightly large in comparison to the others

Additionally, `window-minimize` used an older shape language and color, so that has been unified as well.

In context:
image
image
image


danirabbit (changes requested):
I think this makes the close button too big. Especially apparent here:

Screenshot from 2024-09-10 16 15 51

Seems way larger than the plus icon

PR #1296, opened 2024-07-31
Unify Symbolic Color: Actions

wpkelso:
Some housekeeping tasks that appears much bigger than it actually is. 
- Corrects all symbolic actions icons to use `Silver 900`, and converts a couple of icons to use only fill paths, rather than stroke paths. 
- Shape, fill areas, etc have not been touched. 
- `object-inverse` was updated from 90dpi to 96dpi.

This is intended to be a part of a set of PRs that will touch all similar folders, so if there is a better way to do this please let me know


wpkelso:
I created a weird git situation, recreating to fix said situation

PR #1297, opened 2024-07-31
Unify Symbolic Color: Actions

wpkelso:
First of a series of housekeeping PRs aimed around switching all symbolic icons to use Silver 900 as their base color.

- Switches all icons to use Silver 900 as a base color
- Converts a few icons to using purely fill-paths, rather than stroke-paths
- Updates `object-inverse` to 96dpi from 90dpi.


danirabbit (approved):
This seems okay afaict. It looks like most of the diff here is formatting changes. A lot of these older icons use fewer new lines

PR #1308, opened 2024-08-28
Status: Use chat bubbles for `user-` icons

### LINKED ISSUE #1307: Status: `user-`* icon styles differ between fullcolor and symbolic
### What Happened?

The `user-` prefixed status icon for the fullcolor icons are colored dots. For symbolic, they are speech bubbles.

### Steps to Reproduce

Fullcolor:
user-offline

Symbolic:
user-offline-symbolic

### Expected Behavior

Ideally they should be the same. The [FD.o naming spec](https://specifications.freedesktop.org/icon-naming-spec/latest/) says these are specifically chat icons for the status of the user in chat. But it doesn't say what design should be used.

elementary seems to use the `user-` icons as device/network/general status icons (though Icon Browser lists them as chat icons). `user-away`, [for example](https://github.com/search?q=org%3Aelementary+user-away&type=code), is used for low ink in printer, when a document modified in Code, bluetooth is pairing, or network needs authentication. `user-offline` and `user-available`, [for example](https://github.com/search?q=org%3Aelementary+user-available&type=code), are used more as networking icons (where maybe `network-offline` would be more appropriate, but the FD.o dictates a style for that one which would not fit, and also there is no `network-available` name in FD.o).

So I don't really know which direction to go to make these uniform.

This shows some limitations in the FD.o spec which maybe could be updated to help resolve this sort of thing.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

newhoa:
I used the chat app style as the basis for this one. Hopefully this PR is okay as-is, but I'll add some notes and alternatives below. Let me know if anything needs changing. Thanks!

---

`user-offline` used an `x` emblem on the symbolic icon, so I stuck with that. I tried the "disabled" slash. I thought it was kind of hard to make the chat bubble out since the slash blocks the tail. I'll put it here in case you all would rather go with that, just let me know:

Current:
user-offline

Alternative:
user-offline-x

---

Possible issue, the 16px icons are 16px wide. I tried to do 14px wide, but it ended up either circular looking or the height needed to be reduced such that it just looked too small, especially for adding emblems inside. Personally I like the 16px one a lot more, but if 14px is needed or looks better let me know and I'll rework them.

Current (16px wide):
user-available

Alternatives (14px wide):
user-available-14px
user-available-14px-shorter

---

I did make icons for the non-FD.o ones before realizing they weren't needed. If anyone wants them:
16px:
 
user-typing
user-invisible
user-busy

24px:

user-typing
user-invisible
user-busy

---

Closes #1307


danirabbit (approved):
These look so good! I agree that the "x" on offline looks better than a slash here. Nice work!

PR #1310, opened 2024-09-10
Devices: add scroll wheel to mouse

danirabbit:
Another idea of making the mouse icon look less like the Magic Mouse while still keeping a modern design more inspired by the Logitech Pebble mouse

Screenshot from 2024-09-10 12 54 58


wpkelso:
I'm inclined to continue with this design path rather than go further with #1271

danirabbit:
@wpkelso I invited you to our UX team so you can push to this branch if you'd like :)

devices/24/input-mouse.svg
wpkelso, line 1: Proportionality was preserved rather than precisely pixel-fitting the scroll wheel; It is fit to a half-pixel grid instead. I don't think it's large enough to be noticeable unless you're specifically looking for it, but it might show up on hiDPI displays

danirabbit (commented):
Can’t approve my own PR but I would be happy to go ahead and merge these :)

wpkelso (approved):

PR #1312, opened 2024-09-18
Actions: Make mail-message-new uniform, add 16px

newhoa:
`mail-message-new` symbolic did not match the 24px version (which was updated more recently in 367c69b). So updated the symbolic icon to match.

Also added a 16px version.

---

I worked on a 16px `mail-send-receive` to match the [symbolic version](https://github.com/elementary/icons/blob/main/actions/symbolic/mail-send-receive-symbolic.svg). Wasn't sure whether to include it in this PR. If it's wanted I can add it here or in another PR, just let me know.

Currently, no fullcolor version of this icon exists. I tried 3 options:

1) Orange and Green based on [this older version](https://github.com/elementary/icons/blob/b4708cf4fc99a202582810941b1a167ac9cba11b/old/elementary/actions/24/mail-send-receive_alt.svg), which elementary-xfce still uses.
2) Purple and Green based on document-send and browser-download
3) Purple and Blue based on import, export (although mail does already use these colors for forward and reply)

mail-send-receive mail-send-receive-send-download mail-send-receive

mail-send-receive-prop

Anyway, let me know if you all want this one and if so what version and whether to include it in this PR or another.

Thanks!


danirabbit (approved):
Thanks for keeping these consistent!

And yeah totally interested in the color mail-send-receive! I think the purple and green one to be consistent with upload/download looks really good personally!

PR #1313, opened 2024-09-19
Actions: Add color mail-send-receive

newhoa:
Options here! I submitted here Option A. But when I went to adjust the arrow aspect ratio of the symbolic icon (which was a little narrow), realized it used longer arrow tails (Option C).

So, Option A is what I submitted, the arrow tail matches the ones used in `document-send` and `browser-download`. Option C, the arrow tail matches `go-up` and `go-down`. Options B and D are just variants.

Just let me know which one is best. Thanks!

Light:

prop1
prop2
prop3

Dark:

dark-prop1
dark-prop2

### DIFF
--- a/actions/16/mail-send-receive.svg
+++ b/actions/16/mail-send-receive.svg
@@ -1,0 +1,200 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   height="16"
+   width="16"
+   version="1.1"
+   id="svg6"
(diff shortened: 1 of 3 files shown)


danirabbit:
I think I prefer C personally. A seems a bit squished to me for some reason

newhoa:
> I think I prefer C personally. A seems a bit squished to me for some reason

Okay, thanks! Went with C and force pushed it. Should be good!

danirabbit (approved):
Thanks so much!

PR #1314, opened 2024-09-23
Places: Refresh mailbox icons

newhoa:
- Use green and purple arrows for `mailbox-` inbox and outbox, consistent with the recently updated colors to mail send/receive (also matches document send, browser download colors and new rounded-arrow style).
- Use slightly more front-facing perspective.
- Add 24px icons. Honestly not sure if the 24px ones are used anywhere, but I was going to add them to elementary-xfce so thought I'd put them in here too. If they aren't needed I can remove them.
- Fixed an issue from last mail PR, removed leftover arrow layer from `mail-send-receive-symbolic`.

[Symbolic icons](https://github.com/elementary/icons/blob/main/places/symbolic/mail-inbox-symbolic.svg) use the more top-down perspective. Is this okay? I couldn't come up with a better design there and think they look nice and clear as-is, even if it is a little inconsistent.

Any feedback would be appreciated. Thanks!

Using Evolution as an example:

Current:
current

Proposed:
propPR2


newhoa:
Played with the perspective of the 24px one a little more, let me know if these alternatives are preferred and I can update the PR:

Current PR:
mail-inbox
mail-mailbox
mail-outbox


Alternatives:
mail-inbox00
mail-mailbox00
mail-outbox00

---

16px for reference:
mail-inbox
mail-mailbox
mail-outbox

teamcons:
> Played with the perspective of the 24px one a little more, let me know if these alternatives are preferred and I can update the PR:

Neutral about it. The difference is too minimal to have an opinion.

danirabbit:
Hm I don't think I really love them being more front-facing tbh. Especially for the empty mailbox it leaves the majority of the canvas really empty

The arrow color and shape change though is of course great and the darker interior of the box also makes the colors pop a bit more and gives the mailbox shape more definition so I like that change too

danirabbit:
@newhoa what do you think about something more top-down using the same shape as the mail app icon? https://github.com/elementary/icons/tree/danirabbit/places-mailbox

PR #1319, opened 2024-09-26
Emotes: Fix symbolic pagesize, grid alignment

newhoa:
The pagesize on these were incorrect (16.01, 15.9, etc as opposed to 16x16). This caused some clipping or alignment issues. Fixing this means there should be less clipping, and a little more sharpness.

See https://github.com/elementary/icons/pull/1315 for more details


danirabbit (approved):
Oh wow thanks for noticing and fixing these!

PR #1322, opened 2024-10-07
Cursors: Update `text` and `text-vertical` cursors

wpkelso:
Targets:
- Updates `text` and `text-vertical` to use newer practices for pointer drop-shadow
- Updates shape language for both pointers

For the initial PR, I have two options ready:

text-cursors-scale

I would appreciate feedback about which option to fully move forward with.


danirabbit (changes requested):
I think it needs a bit deeper of a divot on the ends. When viewed at the regular cursor size on LoDPI, the divot isn't really perceptible as compared to the old version

Also, these appear to be a different size than the old cursors. You'll need to update the .in files that go with these to make sure they stay centered, or make sure they are the same size as the old ones

I don't think I've seen a version of this cursor with a crossbar before. The version without is probably better

wpkelso:
> I don't think I've seen a version of this cursor with a crossbar before

I guess I could've included it in a 'prior art section', but MacOS/Apple/AppKit [uses a crossbar on their iBeam](developer.apple.com/documentation/appkit/nscursor). As far as I can tell, that's the only major platform that uses a crossbar.

I'll push those corrections to the barless version in a bit.

wpkelso:
Current state of things

image

danirabbit (changes requested):
Don't forget you either need to resize these so they're the same size as the old ones were or update the associated .in files with new coordinates

wpkelso:
Dev environment is misbehaving, so I haven't been able to completely test, but the `text` hotspot should be fixed now. I don't think `text-vertical` needed any change in its hotspot, but again, I haven't been able to do more than a visual "that looks right."

I'll do more thorough testing if I can get my environment to behave.

danirabbit (approved):
LGTM. Nice work!

PR #1338, opened 2025-02-06
Places: redesign folder-downloads

danirabbit:
Based on the browser-download shape. In context:

Screenshot from 2025-02-06 09 33 49


teamcons:
I really like this one.
The cloud is a more lively design than a circle, and more evocative of internet/the cloud.
Would it work for very small sizes with the same green as the circle we have one ?

PR #1339, opened 2025-02-06
Places: add start-here

danirabbit:
Add `start-here` as defined in the spec: https://specifications.freedesktop.org/icon-naming-spec/latest/#names

using a link to edit-find because that's what we use in the applications-menu by default


wpkelso (approved):
Makes sense to me 👍

PR #1342, opened 2025-03-15
Places: Create symlinks in Meson

danirabbit:
Create links in the build system. Towards https://github.com/elementary/icons/issues/578

* `network-workgroup` is a link because it has a unique icon in symbolic size
* `folder-remote` and `folder-publicshare` are links at symbolic­­-24px because they have unique icons at 32px and above
* Also fixes an issue with broken `start-here` symlinks.


wpkelso (approved):
I see no issues with this

PR #1346, opened 2025-04-04
Mimes: Copy base files for templates at build time

wpkelso:
Extends the work in #1334 to copy base files to template files, then add the decoration all at the build time, giving us greater consistency.


danirabbit (approved):
Nice! Good cleanup

Issue #984, opened 2020-12-28
Inkscape as a flatpak can't read palettes 


danirabbit:
## What Happened

When installing Inkscape as a user flatpak, it can't read the palette files

## Expected Behavior

We should make sure we install palette files wherever they need to be so that Inkscape in a flatpak has permission to read them

wpkelso:
As a long-term solution, perhaps including the elementary palette upstream in Inkscape itself would be something to look into. Ubuntu, Gnome, and Android all have palettes included in the default palette list, so I don't think including elementary's would be completely out of the question

PR #1349, opened 2025-05-29
Prep palette file for Inkscape

wpkelso:
I proposed getting the palette shipped with [Inkscape](https://gitlab.com/inkscape/inkscape/-/tree/master/share/palettes?ref_type=heads) in #984 as a longer term solution, and this is a follow-up to make sure we're all prim and proper before pursuing any such merge requests.


danirabbit (approved):
:rocket:

PR #1352, opened 2025-06-12
Devices: differentiate keyboard from keyboard preferences

danirabbit:
Make the devices icon look like its color counterparts instead of it looking like the keyboard preferences icon.

Context:

## BEFORE

Screenshot from 2025-06-11 18 06 14

## AFTER

Screenshot from 2025-06-11 18 07 51


wpkelso (approved):
Very cute! 🚀

Issue #1353, opened 2025-06-18
Remove drop shadow from app icons


danirabbit:
We’ve previously applied a “shelf” style shadow to app icons that fits in with a shelf style dock. However, we no longer have this style in the dock.

Additionally other freedesktop platforms like GNOME are recommending app authors not to apply drop shadow to their app icons at all, and they handle app icon shadows on the shell so that they’re applied consistently and scaled uniformly. This means that many third party icons now have no drop shadows and look out of place in Pantheon

We should probably go ahead and remove drop shadows from our app icons as well and do shadows in the shell so that we don’t end up with a mix of app icons with shadows and some without

PR #1362, opened 2025-07-10
Apps: Remove shadows and clean up rim-lights

wpkelso:
Rel #1353 

Cleaned up some erroneous rim-light stops while we were here


wpkelso:
> I can't remember if we've already broken anything


Looks like icons were removed in 8297fdd09bdc5a20715df4de5e9f653b04c66fc3 and a8e06ebc4defc102ce136321feacea54d27916a2

PR #1365, opened 2025-08-11
Use Adwaita as a fallback icon theme

daudix:
Currently, if no fitting icon is found in elementary icon theme, gtk falls back to `hicolor`, as per the fd.o spec. With this PR, it will first try to find missing icon in Adwaita (which is present in elementary OS) before resorting to displaying a missing icon image. This issue mainly affects GNOME Circle apps that heavily rely on Adwaita, and while it is recommended to bundle all non-fd.o icons, not all apps do so (unfortunately).

Before/after examples (Apostrophe and Ptyxis):

Screenshot from 2025-08-11 22 07 16
Screenshot from 2025-08-11 22 06 50

Screenshot from 2025-08-11 22 27 56
Screenshot from 2025-08-11 22 28 12

### DIFF
--- a/index.theme.in
+++ b/index.theme.in
@@ -1,6 +1,7 @@
 [Icon Theme]
 Name=elementary
 Comment=Smooth modern theme designed to be intuitive.
+Inherits=Adwaita
 
 Example=directory-x-normal
 


danirabbit (approved):
As far as I can tell this works as expected and doesn't override icons gresourced in granite

tintou:
Every time that I report it, the app get fixed in a few days, I'm not sure that we are doing anything good with the ecosystem to apply this workaround just for us

danirabbit:
@tintou hm yeah maybe you're right. If you want to add a PR to revert with that reasoning, I guess I'll revert. And we should probably add a comment to the theme file about why we don't have `inherits`.

PR #1367, opened 2025-08-15
Remove visual artifact from 16px edit-paste icon

choosingausernameishard:
The paste icon had this weird shadow I assume was unintentional:


wpkelso (approved):
Good catch! Thanks ❤️

PR #1372, opened 2025-08-16
Redesign changes-allow

choosingausernameishard:
This is the third and final pull request split from #1369. It contains the icon I submitted as `unlocked.svg`in #1369, but this time as a redesign of `changes-allow`, which I wasn’t aware already existed.

The new design has the advantage of being more visually distinct from the `locked` icon, especially at small sizes.


danirabbit (approved):
Gonna go ahead and merge this since we have some broken symlinks currently in main and we can always revisit this later

PR #1373, opened 2025-08-17
Add 16px preferences-system.svg

choosingausernameishard:
It is used in Inkscape as the icon for Inkscape’s preferences


danirabbit (approved):
🚀

PR #1374, opened 2025-08-17
Add 16px document-properties

choosingausernameishard:
It is used in Inkscape


danirabbit (approved):
Looks really good nice work!

PR #1383, opened 2025-09-01
Add dialog-text-and-font for Inkscape

ghost:
Screenshot 2025-09-01 at 12 05 49


danirabbit (commented):
Seems like this should probably be a symlink to `preferences-desktop-font`?

PR #1368, opened 2025-08-15
Inkscape tool icons

choosingausernameishard:
Added the missing Inkscape graphics tool icons as well as 16px versions of existing icons (because the icon size can be customised by the user).

Here is a screenshot of the Inkscape vertical toolbar with 24px icons:

Screenshot 2025-08-14 at 23 11 39

And below with 16px icons:

Screenshot 2025-08-14 at 23 13 02

PR #911, opened 2020-07-04
Add Inkscape's  Distributions Icons

Fatih20:
Fixes #763 and then some. It also add `distribute-graph and symbolic icons of them.

This is how the icons look under elementary's light theme
image
This is how the icons look under elementary's dark theme
image
This is how the symbolic icons look under elementary's light theme
image
This is how the symbolic icons look under elementary's dark theme
image

There is some things that is still in the file that I use in the design process of these icons. I'll be cleaning them up once these are finalized and approved.

PR #1385, opened 2025-09-01
Add dialog-svg-font for Inkscape

choosingausernameishard:
Screenshot 2025-09-01 at 12 15 51

For Inkscape’s “SVG Font Editor”. The orange asterisk is the logo for SVG.

This icon was more difficult to design, here are some of my previous attempts:

  

Let me know if you prefer one of those instead.


choosingausernameishard:
@wpkelso I’ve switched it to that version. Could you please also take a look at #1368, in which I add icons for all of Inkscape’s tools, and also at #911, in which @Fatih20 adds alignment-related icons? Thank you!

wpkelso (approved):
🚀

PR #1390, opened 2025-09-01
Add bitmap-trace for Inkscape

choosingausernameishard:
The icon is based on `image-x-generic`.

Screenshot 2025-09-01 at 18 00 54


wpkelso (approved):
I think this works

PR #1394, opened 2025-10-31
Release 8.2.0

danirabbit:
Need this before we can release Dock with workspace switcher.

The removals are non-fd.o icons that had really specific uses and are already gresourced in the old places they were used


micahilbery (approved):
🚀

PR #1405, opened 2025-12-16
Add edit-duplicate for Inkscape

ghost:
I based it on `edit-copy`. Though it’s intended for Inkscape, the icon’s generic name means it’s likely to be used by some other apps as well, so I designed it in all of `edit-copy`’s sizes.


wpkelso (commented):
~~Where is this used? I don't see it showing up anywhere in menus or toolbars.~~ Does this need to be a separate icon? It seems like it would be sufficient to just have a symlink to `edit-copy`, given the action is essentially a macro for copy + paste in place.

Given this isn't an [fd.o](https://specifications.freedesktop.org/icon-naming/latest/) icon name, we don't need to worry about non-Inkscape uses here. However, I agree it is a fairly generic icon, so it might be a good candidate to propose to add to Granite.

PR #1415, opened 2026-01-13
Devices: Add pointing stick

wpkelso:
Rel https://github.com/elementary/settings-mouse-touchpad/pull/263


tintou:
We should ask for inclusion in the fdo spec

PR #1419, opened 2026-03-24
Add New Red Calendar Icon

kurisubrooks:
This pull request refreshes the existing `apps/office-calendar` icons, which currently features a green calendar pad with a grid, with a red, spiral bound pad, and dots to represent the days of the week. There's also a large red "x" to denote the day of an event in the 2nd-to-last column (or last column at low resolutions).

I've also updated the `actions/events-new` icons with the new red calendar style. I also removed the red "x" mark from this one since the icon has a "+" emblem, and that could potentially confuse users.

Both `office-calendar` and `events-new` icons were also fairly out-of-date in this repository, since the `calendar` icon in the `elementary/calendar` repo was updated 4 years ago with the new base square shape (increased corner radius). 

Since these proposed icons were based on the newer shape from this repo, these icons should be more consistent with the current/more modern style of elementary as a result.

> This pull request will be paired with a separate pull request (to be made in [elementary/calendar](https://github.com/elementary/calendar)) to update the icon of the `io.elementary.calendar` app.

### Preview
Dock

In editor (hi-res)

In editor (1x scale, pixel-accurate preview)


danirabbit:
Hey thanks for your pull request!

✅ Definitely appreciate updating to the new base icon shape and corner radius etc here. Thank you!

✅ We used to have a red calendar icon and we moved away from it, I think just to try to do something unique. But I'm not personally super attached to the green. It seems other calendar apps have converged around the red so I'm happy to go back to red here.

❌ We also used to have the spirals and I'm personally not a fan of this detail. I don't really feel like this adds to the icon and I'd prefer not to bring that back. (See commit history: https://github.com/elementary/icons/commits/main/apps/48/office-calendar.svg)

❌ I'm also not totally sold on the "x" detail. Especially since as you noted it needs to be removed for variants. But it also feels to "action"-y to me when app icons should feel more "noun"-y. It seems like it's there to convey information but it actually doesn't. I don't think this detail is an improvement

🟡 The dots are interesting as an alternative to the grid, but I don't think I love how you scaled them here for 24px. I think I'd prefer to see smaller dots and keep the three rows personally.

🟡 I also like to see chromatic grays instead if we can. For the dots, maybe instead of 40% of black try like 50% of Cocoa 900. 

🟡 I think the highlights under the dots are a little too intense and it's giving that kind of "pegboard" effect. Maybe those should be a bit more transparent

🟡 It seems like you're interested in breaking up the tile shape, what about something like a corner page flip? 

pngtree-calendar-vector-art-icon-png-image_8076673-4015868535

🟡 `event-new` is actually not an fd.o icon name and it looks like we already resource that in Calendar, so I wouldn't mind if you removed it here altogether while updating these since we'll want to remove it anyways.

PR #1421, opened 2026-05-03
Places: update symbolic icons for gtk 4.22

garaevdi:
Part one of the grand symbolic icons update for gtk 4.22

 - With inkscape, remove any groups and "bake" transforms into the paths by moving them back and forth
 - Using gtk4-icon-editor change icon's color to the appropriate color name
 - With a code editor (in my case zed) find and replace fallback colors from black to silver 900


danirabbit (approved):
afaict these all work fine with older GTK as well. Nice work! Thanks for the cleanup

PR #1422, opened 2026-05-06
Status: update symbolic icons for gtk 4.22

garaevdi:
Second part of grand update of symbolic icons. Finally had some time to test out second biggest group of icons with gtk3 (Is there something like Icon Library but for gtk3? `gtk3-icon-browser` only shows some icons).

Pretty much the same as #1421


stsdc:
Hi @garaevdi, did you check https://github.com/elementary/iconbrowser?

status/24/battery-missing-symbolic.svg
```diff
@@ -1,45 +1,66 @@
...
-  <defs
-     id="defs12" />
+        xmlns='http://www.w3.org/2000/svg'
+        xmlns:svg='http://www.w3.org/2000/svg'
+        id='svg8'
+        width='24'
+        height='24'>
   <path
-     id="path648"
-     class="error"
```
danirabbit, line 27: It looks like all the style classes were removed. Can we keep these for compatibility with older GTK?

Image

danirabbit (changes requested):

status/24/battery-missing-symbolic.svg
garaevdi: Could you test it again? It seems like gtk 4.14 doesn't really know how to apply multiple style classes with single quotes. I naively assumed that if gtk3 can handle this, then it would be a piece of cake for gtk4.

garaevdi:
> Hi @garaevdi, did you check https://github.com/elementary/iconbrowser?

Hi, I thought it was gtk4 since the beginning. Tested out last commit before the switch and it works fine, thanks!

status/24/battery-missing-symbolic.svg
danirabbit: looks good now! Thanks :)

danirabbit (approved):
Nice work! Thanks

PR #1434, opened 2026-08-12
Actions: remove non-fd.o event-new

danirabbit:
This is non-fd.o and we already gresource it in Calendar


lenemter:
This icon is present in Icon Browser. You need to remove it from there as well

wpkelso (approved):
I'm going to hold merging this until we can do it concurrent with a removal from icon browser

danirabbit:
fwiw iconbrowser will just skip missing icons, but https://github.com/elementary/iconbrowser/pull/78

PR #1435, opened 2026-08-12
Apps: remove non-fd.o office-calendar

danirabbit:
`office-calendar` isn't an fd.o app icon name. We already gresource app icons in Calendar and we resource `office-calendar-symbolic` in Granite so the apps that use it still have access


wpkelso (changes requested):
It looks like `x-office-calendar` is in the fd.o mimetypes, so we need to keep the ones in `mimes/` around

https://specifications.freedesktop.org/icon-naming/latest/#mimetypes

danirabbit:
@wpkelso yeah I left those! They were symlinks so I moved the old ones

wpkelso:
@danirabbit Oh crap, you're right, I'm sorry. I completely misread the diff!

wpkelso (approved):

Issue #1420, opened 2026-04-17
Broken symbolics on Gtk 4.22 (symbolic format VS Transforms)


teamcons:
### What Happened?

Several emoji icons do not follow Gtk symbolic icon format as described: https://docs.gtk.org/gtk4/icon-format.html
Apparently there are transforms which arent supposed to go in as part of limitations, and thus arent rendered in Gtk 4.22. It may work normally in 4.24

Some affected:
> emblem-default-symbolic
> list-add-symbolic
> mail-send-symbolic
> dialog-password-symbolic
> face-smile-symbolic
> face-smile-big-symbolic
...(More emojis)

<img width="344" height="376" alt="Image" src="https://github.com/user-attachments/assets/8bd989e6-a887-483e-ab27-1baa84472025" />

Related: 
[elementary symbolic icons not rendering when app uses Gnome Platform 50](https://github.com/flathub/io.elementary.BaseApp/issues/48#issue-4185503863)

[GTK issue report of elementary icons not rendering on Gtk 4.22](https://gitlab.gnome.org/GNOME/gtk/-/issues/8147)

[discussion about Gtk 4.21 and broken icons](https://github.com/elementary/icons/discussions/1417#)

### Steps to Reproduce

1. ...[Download the latest release of Jorts](https://github.com/elly-code/jorts/releases/tag/4.0.1-9)
2. ...bump the Gnome Runtime to 50 in the io.github.elly_code.jorts.devel.yml flatpak manifest
3. `flatpak run org.flatpak.Builder --force-clean --sandbox --user --install --install-deps-from=flathub --ccache --mirror-screenshots-url=https://dl.flathub.org/media/ --repo=repo builddir io.github.elly_code.jorts.devel.yml`

### Expected Behavior

icons render perfectly fine

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

ryonakano:
@garaevdi Sorry to ping you, may I ask you the current progress for this?

PR #1440, opened 2026-08-19
metainfo: Add release note for 9.0.0

ryonakano:
I personally would like to release a new version of icons to introduce the fixes for #1420 into our flatpak-platform. I don't maintain this repository (i.e. I have had few contributions) so let me know if you have some other fixes that should be in 9.0.0.

### DIFF
--- a/data/icons.metainfo.xml.in
+++ b/data/icons.metainfo.xml.in
@@ -29,7 +29,7 @@
   <update_contact>contact_at_elementary.io</update_contact>
 
   <releases>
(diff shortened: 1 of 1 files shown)


danirabbit:
I think we should bump to 9 probably because we have removals so we want to make sure we're not building this for OS 8

ryonakano:
Updated the version in the release note. Also restricted the distribution series in Launchpad: https://code.launchpad.net/~elementary-os/+recipe/icons-daily

danirabbit (approved):

PR #1444, opened 2026-08-29
Add `adw-entry-edit-symbolic` symlink

ztefn:
Same as elementary/granite#1003 but for plain libadwaita apps.


danirabbit:
This should be a line in granite's icon resource instead of a symlink in this repo. Eventually all symbolic icons will move to Granite since it's out of spec 

Adw should resource icons that it depends on

